PR #2, opened 2020-05-04
Add observed generation and samples.

jortel:
Add support for _ObservedGeneration_.

Assuming that:
1. The controller will only run on Kubernetes 1.10+
2. We can avoid the need for the controller to ever update the `Spec`.

we can leverage the `Status` _subresource_ which simplifies things a bit.
For example, the status update does not cause the `Generation` to be incremented so the ObservedGeneration can simply be set equal  to the Generation.


djwhatle:
@jortel cool usage of status subresource. I guess because this controller doesn't need to work on OCP 3.x we can use it safely.

PR #20, opened 2020-08-31
Fix tree builder.

jortel:
Fixed a flaw in the `Tree` builder when the leaf is a `Folder`.
Discovered this while testing a new folder tree endpoint that I then determined that I did not need.

### DIFF
--- a/pkg/controller/provider/model/vsphere/tree.go
+++ b/pkg/controller/provider/model/vsphere/tree.go
@@ -19,7 +19,14 @@ var (
 	VmKind         = libref.ToKind(VM{})
 )
 
-var InvalidRefErr = liberr.New("invalid ref")
+//
+// Errors
+var (
+	// Invalid ref.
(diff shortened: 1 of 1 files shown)


pkg/controller/provider/model/vsphere/tree.go
```diff
@@ -79,7 +86,14 @@ func (r *Tree) Build() (*TreeNode, error) {
 			folder := model.(*Folder)
 			refList := RefList{}
 			refList.With(folder.Children)
+		next:
 			for _, ref := range refList {
```
mansam, line 90: I only see one `for` here, is the label on the continue necessary?

jortel: Yes because a _continue_ (or break) in Go applies to the inner most loop/switch so a _naked_ continue would only apply to the switch but I needed it to apply to the `for` loop.

PR #21, opened 2020-09-01
Split resource map CRs and controllers.

jortel:
The UI (and user) want to manage resoruce map templates for networks and storage separately.  Also, we'd always planned for the templates to be copied into the plan instead of being referenced.

The `map` CR and controller split into 2:
map/network
map/storage

And included the maps in the plan CR.


config/crds/virt_v1alpha1_storagemap.yaml
```diff
@@ -0,0 +1,81 @@
...
+                        description: A storage class.
+                        type: string
+                    required:
+                    - storageClass
+                    type: object
+                  source:
+                    description: Source storage.
+                    properties:
+                      id:
+                        description: 'The network identifier. For:   - vsphere: The
```
mansam, line 50: Network should be storage

config/crds/virt_v1alpha1_plan.yaml
```diff
@@ -32,6 +32,71 @@ spec:
...
+                            description: A storage class.
+                            type: string
+                        required:
+                        - storageClass
+                        type: object
+                      source:
+                        description: Source storage.
+                        properties:
+                          id:
+                            description: 'The network identifier. For:   - vsphere:
```
mansam, line 53: This should be 'storage' rather than 'network'

config/crds/virt_v1alpha1_storagemap.yaml
jortel: Strange, I don't think this matches the common on the struct.

jortel: Likely due to anonymous struct.  Will refactor with named struct.  This lends itself to the _source_ being reused for both network and storage so will make the comment more generic.
Thinking the anonymous struct may not be a good idea.  Also, found that the field being an anonymous struct make is hard (impossible) to fix the unit tests.

PR #22, opened 2020-09-03
Provider: inventory OCP resources.

jortel:
Add OCP resources to the inventory for networks and storage.

The `.../provider/openshift/<name>/networkattachmentdefinitinitions` endpoint sure seems long but being explicit has been the policy which guess is best.


mansam (approved):
LGTM

PR #26, opened 2020-09-15
Next iteration of valiations.

jortel:
Update network & storage map CRs to reference a _pair_ of source/destination providers.

Create new `validation` package under controller to support validating shared CR data structures:
- Provider (ref)
- ProviderPair (ref)
- NetworkPair
- StoragePair

Added `description` to the Plan CR.  Needed by the UI.
The _Message constants were, well, messy and since conditions only created in one place, I got rid of the __Message const.

Requires: https://github.com/konveyor/controller/pull/24

### DIFF
--- a/pkg/controller/provider/web/ocp/client.go
+++ b/pkg/controller/provider/web/ocp/client.go
@@ -114,6 +114,28 @@ func (c *Client) Path(object interface{}, id string) (path string, err error) {
 				Name:      name,
 			},
 		})
+	case *StorageClass:
+		h := StorageClassHandler{}
+		path = h.Link(
+			&c.Provider,
+			&model.StorageClass{
+				Base: model.Base{
+					Name: name,
+				},
+			})
+	case *NetworkAttachmentDefinition:
+		if id == "" { // list
+			ns = c.Provider.Namespace
+		}
+		h := NetworkAttachmentDefinitionHandler{}
+		path = h.Link(
+			&c.Provider,
+			&model.NetworkAttachmentDefinition{
+				Base: model.Base{
+					Namespace: ns,
+					Name:      name,
+				},
+			})
 	default:
 		err = base.ResourceNotSupported
 	}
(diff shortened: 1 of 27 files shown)


pkg/controller/provider/web/ocp/client.go
```diff
@@ -114,6 +114,31 @@ func (c *Client) Path(object interface{}, id string) (path string, err error) {
 				Name:      name,
 			},
 		})
+	case *StorageClass:
+		if id == "" { // list
```
mansam, line 118: is this conditional doing anything, since storageclasses don't have namespaces and `ns` doesn't seem to be used in the rest of the block?

pkg/controller/validation/network.go
```diff
@@ -0,0 +1,189 @@
...
+	switch provider.Type() {
+	case api.OpenShift:
+		resource = &ocp.NetworkAttachmentDefinition{}
+	case api.VSphere:
+		return
+	default:
+		err = web.ProviderNotSupported
+		return
+	}
+next:
```
mansam, line 136: I don't think this is necessary, because as I understand it `continue` doesn't interact with switch statements in Go; it should just continue to the next element in the surrounding `for`. Check out this example: https://play.golang.org/p/GSpIJhoVD2S

mansam (approved):
Looks great, just a couple minor comments/questions.

pkg/controller/validation/network.go
jortel: Yes, it is.  The continue will apply to the switch.  See [here](https://play.golang.org/p/Dqg274HmwXY) I added a print after the switch.

pkg/controller/provider/web/ocp/client.go
jortel: Ah, you're right.  cut-n-pasted and it's not needed.

jortel:
Rebased on [controller/pull/24](https://github.com/konveyor/controller/pull/24) and squashed.

PR #28, opened 2020-09-25
Run migrations - part 2

jortel:
Moved the itinerary logic and per-VM status tracking to the plan controller.
Started building the VirtualMachineImport CR.


mansam (approved):
LGTM

PR #30, opened 2020-09-29
API client improved.

jortel:
This de-duplicates a good bit of code in the `web/ocp` and `web/vsphere` packages by introducing the convent of a `Resolver`.
The _Resolver_ resolves a REST resource object (kind) into a URL path.  Resolving the path **was** the main reason that the _type-sepecific_ package had to implement their own Client.

The provider API _Ready_ functionality has been centralized in the `ProviderClient` and rather than being called specifically, the provider _readiness_ is check on 404.  This also elliminates race conditions between the client making the GET call and the provider controller add/delete from inventory.

This PR also reduces complexity everywhere the provider REST API client is used in the other controllers because they no longer need to check for 206 _PartialContent_ because the `ProviderClient` raises the appropriate error.


mansam (approved):
Great stuff @jortel

PR #33, opened 2020-10-12
Add Networks and Concerns to VM REST resource.

jortel:
Add Networks and Concerns to VM REST resource.
The _concerns_ will eventually be populated by integration with MA shared validation service.


mansam (approved):
:+1:

PR #51, opened 2020-10-23
Better support for hosts.

jortel:
Add `Secret` ref to `Host` CR.
Add `Host` validations.  Refactor to use shared provider validation.
Update controller flow to find Hosts in the same namespace and use the secret instead of the provider secret when found.  The logic also changed to create a Secret for each VirtualMachineImport CR.  Ideally we could create only the secrets needed but saving that for a future optimization.
Add VM.Host to the provider inventory.
Renamed and standardized variable for the provider API (as inventory) to clarify between that client and the k8s client.

Added `Referenced` object to be included in our CRs used to collect instances of resources fetched by validation to be used further downstream in reconcile logic.  The intent is to replace the stashing of resources in the Plan annotations in a follow up PR.

### DIFF
--- a/config/crds/virt_v1alpha1_host.yaml
+++ b/config/crds/virt_v1alpha1_host.yaml
@@ -40,6 +40,9 @@ spec:
             provider:
               description: Provider
               type: object
+            secret:
+              description: Credentials.
+              type: object
           required:
           - provider
           - id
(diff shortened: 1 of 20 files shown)


mansam (approved):
Looks great

PR #55, opened 2020-10-27
Replace plan snapshot with Referenced.

jortel:
Replace plan snapshot with Referenced.
Also, noticed while troubleshooting that storing the validation result before running the migration would be helpful.  This should be safe since the Running, Succeeded, Failed conditions are durable.  Otherwise, if an error occurs while running the migration, the Plan will not reflect `Ready`.

### DIFF
--- a/pkg/apis/virt/v1alpha1/snapshot/annotation.go
+++ b/pkg/apis/virt/v1alpha1/snapshot/annotation.go
@@ -85,7 +85,7 @@ func (r *Annotation) Get(key string, object interface{}) (err error) {
 //
 // Contains key.
 func (r *Annotation) Contains(key string) (found bool) {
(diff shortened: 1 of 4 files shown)


pkg/apis/virt/v1alpha1/snapshot/annotation.go
```diff
@@ -85,7 +85,7 @@ func (r *Annotation) Get(key string, object interface{}) (err error) {
 //
 // Contains key.
 func (r *Annotation) Contains(key string) (found bool) {
-	_, found = r.GetAnnotations()[key]
+	_, found = r.GetAnnotations()[r.key(key)]
```
jortel, line 88: This was broken.

mansam (approved):

PR #56, opened 2020-10-28
Persist the VM name after migration

fdupont-redhat:
This pull request persists the VM name after migration. It simply sets the `targetVmName` attribute of the `VirtualMachineImport` CR to the name of the source VM. Otherwise, the name of the VM will be the operating system followed be a random string, which is not user friendly.


fdupont-redhat:
Closed in favor of https://github.com/konveyor/virt-controller/pull/58.

PR #57, opened 2020-10-28
Update plan to reflect progress as needed in execute().

jortel:
Moved the plan update (to reflect progress) to execute() so the plan is not double updated unnecessarily.


pkg/controller/plan/builder/vsphere/builder.go
```diff
@@ -45,8 +45,8 @@ func (r *Builder) Secret(vmID string, in, object *core.Secret) (err error) {
 	if hostFound {
 		hostURL := liburl.URL{
 			Scheme: "https",
-			Host: host.Spec.IpAddress,
-			Path: vim25.Path,
+			Host:   host.Spec.IpAddress,
+			Path:   vim25.Path,
```
jortel, line 49: This seems to just be a `gofmt` that did not get included in an preceding PR.

mansam (approved):

PR #58, opened 2020-10-28
Propagate VM target name.

jortel:
Propagate VM target name from the source provider.
Given the need for provider _type_ content in the VirtualMachineImportSpec, I think it makes sense to rename the Builder.Source() to Builder.Import() and broaden its scope.  This would be better than adding an additional method to the `Builder` interface because it's simpler and supports more future needs.

### DIFF
--- a/pkg/controller/plan/builder/vsphere/builder.go
+++ b/pkg/controller/plan/builder/vsphere/builder.go
@@ -45,8 +45,8 @@ func (r *Builder) Secret(vmID string, in, object *core.Secret) (err error) {
 	if hostFound {
 		hostURL := liburl.URL{
 			Scheme: "https",
-			Host: host.Spec.IpAddress,
-			Path: vim25.Path,
+			Host:   host.Spec.IpAddress,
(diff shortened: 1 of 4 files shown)


pkg/controller/plan/builder/vsphere/builder.go
```diff
@@ -45,8 +45,8 @@ func (r *Builder) Secret(vmID string, in, object *core.Secret) (err error) {
 	if hostFound {
 		hostURL := liburl.URL{
 			Scheme: "https",
-			Host: host.Spec.IpAddress,
-			Path: vim25.Path,
+			Host:   host.Spec.IpAddress,
```
fdupont-redhat, line 48: Has anything really changed on these lines?

jortel: No.  The makefile runs `gofmt` and looks like these lines changed in an earlier PR that got merged without a final `gofmt`.

fdupont-redhat (approved):
LGTM

PR #64, opened 2020-10-30
Add thumbprint to Host model

fdupont-redhat:
This pull request adds the `thumbprint` attribute to the Host model.
The data is collected from VMware via `MOR:HostSystem.summary.config.sslThumbprint`.

### DIFF
--- a/pkg/controller/provider/web/vsphere/host.go
+++ b/pkg/controller/provider/web/vsphere/host.go
@@ -115,7 +115,7 @@ type Host struct {
 	InMaintenanceMode bool          `json:"inMaintenance"`
 	ProductName       string        `json:"productName"`
 	ProductVersion    string        `json:"productVersion"`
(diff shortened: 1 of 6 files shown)


pkg/controller/provider/web/vsphere/host.go
```diff
@@ -115,6 +115,7 @@ type Host struct {
 	InMaintenanceMode bool          `json:"inMaintenance"`
 	ProductName       string        `json:"productName"`
 	ProductVersion    string        `json:"productVersion"`
+	Thumbprint        string        `json:thumbprint`
```
jortel, line 118: The tag is malformed.  The Go tag must be name:"" where the body is in quotes.
Should be:
```
`json:"thumbprint"`
```

jortel (approved):
LGTM

PR #78, opened 2020-11-13
Add integration with MA.

jortel:
Initial integration with MA (xavier) validation service.
Adds:
- _Xavier_ client.
- A `revision` property to all resources.
- A `revisionAnalyzed` property which indicates the `revision` reflected in the analytics `concerns`.
- Triggered by _watch_ on vSphere Cluster and VM database entities.  Likely add a _watch_ on other kinds (Eg: Network) in the next iteration.

Need operator to set:
- XAVIER_URL
- XAVIER_USER
- XAVIER_PASSWORD

 in the virt-controller configMap.  Related PR (TBD).  Re: [issue/43](https://github.com/konveyor/virt-operator/issues/43).

Also, includes /vendor change for _common_ controller lib fix.

Requires: https://github.com/konveyor/controller/pull/34

### DIFF
--- a/pkg/controller/provider/container/ocp/collection.go
+++ b/pkg/controller/provider/container/ocp/collection.go
@@ -53,7 +53,7 @@ func (r *StorageClass) Reconcile(ctx context.Context) (err error) {
 		m.With(&resource)
 		r.Reconciler.UpdateThreshold(m)
 		Log.Info("Create", libref.ToKind(m), m.String())
-		err = db.Insert(m)
+		err = tx.Insert(m)
 		if err != nil {
 			err = liberr.Wrap(err)
 			return
@@ -156,7 +156,7 @@ func (r *NetworkAttachmentDefinition) Reconcile(ctx context.Context) (err error)
 		m.With(&resource)
 		r.Reconciler.UpdateThreshold(m)
 		Log.Info("Create", libref.ToKind(m), m.String())
-		err = db.Insert(m)
+		err = tx.Insert(m)
 		if err != nil {
 			err = liberr.Wrap(err)
 			return
@@ -259,7 +259,7 @@ func (r *Namespace) Reconcile(ctx context.Context) (err error) {
 		m.With(&resource)
 		r.Reconciler.UpdateThreshold(m)
 		Log.Info("Create", libref.ToKind(m), m.String())
-		err = db.Insert(m)
+		err = tx.Insert(m)
 		if err != nil {
 			err = liberr.Wrap(err)
 			return
--- a/pkg/controller/provider/container/vsphere/reconciler.go
+++ b/pkg/controller/provider/container/vsphere/reconciler.go
@@ -349,10 +349,15 @@ func (r *Reconciler) getUpdates(ctx context.Context) error {
 		This:    pc.Reference(),
 		Options: filter.Options,
 	}
-	var transaction *libmodel.Tx
+	var tx *libmodel.Tx
+	watchList := []*libmodel.Watch{}
 	defer func() {
-		if transaction != nil {
-			transaction.End()
+		r.consistent = false
+		for _, w := range watchList {
+			r.db.EndWatch(w)
+		}
+		if tx != nil {
+			tx.End()
 		}
 	}()
 next:
@@ -375,12 +380,12 @@ next:
 			continue next
 		}
 		req.Version = updateSet.Version
-		transaction, err = r.db.Begin()
+		tx, err = r.db.Begin()
 		if err != nil {
 			return liberr.Wrap(err)
 		}
 		for _, fs := range updateSet.FilterSet {
-			err = r.apply(ctx, fs.ObjectSet)
+			err = r.apply(ctx, tx, fs.ObjectSet)
 			if err != nil {
 				err = liberr.Wrap(err)
 				Log.Trace(err)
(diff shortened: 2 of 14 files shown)


pkg/controller/provider/container/ocp/collection.go
```diff
@@ -53,7 +53,7 @@ func (r *StorageClass) Reconcile(ctx context.Context) (err error) {
 		m.With(&resource)
 		r.Reconciler.UpdateThreshold(m)
 		Log.Info("Create", libref.ToKind(m), m.String())
-		err = db.Insert(m)
+		err = tx.Insert(m)
```
jortel, line 56: User transaction per _common_ controller change.

pkg/controller/provider/container/vsphere/reconciler.go
```diff
@@ -375,30 +375,31 @@ next:
 			continue next
 		}
 		req.Version = updateSet.Version
-		transaction, err = r.db.Begin()
```
jortel, line 378: just to be consistent.

pkg/controller/provider/model/vsphere/model.go
```diff
@@ -23,8 +28,8 @@ type Base struct {
 	Name string `sql:""`
 	// Parent
 	Parent string `sql:"index(a)"`
-	// Annotations
-	Annotations string `sql:""`
```
jortel, line 27: Annotations did not end up being generally useful.
Replaced with `concerns` on VM.

pkg/controller/provider/model/vsphere/model.go
```diff
@@ -49,6 +58,26 @@ func (m *Base) Equals(other libmodel.Model) bool {
...
+}
+
+//
+// Updated.
+// Increment revision. Should ONLY be called by
+// the reconciler.
+func (m *Base) Updated() {
+	m.Revision++
+}
+
```
jortel, line 80: Methods for managing the revision.

pkg/controller/provider/xavier/base/client.go
```diff
@@ -0,0 +1,335 @@
...
+		"2",
+		"value",
+		"org.drools.core.runtime.rule.impl.FlatQueryResults",
+		"idResultMaps",
+		"element",
+		"0",
+		"element",
+		"0",
+		"value",
+		"org.jboss.xavier.analytics.pojo.output.workload.inventory.WorkloadInventoryReportModel")
```
jortel, line 124: The path into the returned REST resource is absurd.

pkg/controller/provider/xavier/base/client.go
```diff
@@ -0,0 +1,335 @@
...
+	if len(concerns) == 0 {
+		recommended := false
+		if list, cast := providers.(List); cast {
+			for _, provider := range list {
+				if provider == "OCP" {
+					recommended = true
+					break
+				}
+			}
+		}
```
jortel, line 154: Rules are triggered by _known_ field values.  So if, for example, an unknown guest OS is passed, some of the rules don't fire resulting in `OCP` not being in the recommendTargetsIMS but flagsIMS = nil.  This added as a catch-all.

pkg/controller/provider/xavier/base/client.go
```diff
@@ -0,0 +1,335 @@
...
+			return
+		}
+	}
+
+	return
+}
+
+//
+// Get object at path.
+func (r Client) find(in interface{}, path ...string) (out interface{}, err error) {
```
jortel, line 288: Needed to safely traverse the absurd REST payloads.

pkg/controller/provider/xavier/vsphere/adapter.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	liberr "github.com/konveyor/controller/pkg/error"
+	libmodel "github.com/konveyor/controller/pkg/inventory/model"
+	model "github.com/konveyor/virt-controller/pkg/controller/provider/model/vsphere"
+	"github.com/konveyor/virt-controller/pkg/controller/provider/xavier/base"
+)
+
+//
+// Model adapter.
```
jortel, line 11: Model adapter is the same pattern used in the _container_ package.

pkg/controller/provider/xavier/base/client.go
```diff
@@ -0,0 +1,335 @@
...
+	header.Add(
+		"Authorization",
+		"Basic "+encoded)
+
+	return
+}
+
+//
+// Post validation request.
+func (r *Client) post(in, out interface{}) (status int, err error) {
```
mansam, line 247: I only see this method called in one place; is it necessary to use an interface type for these parameters? It looks like at the very least `out` will always be a `base.Object{}`.

pkg/controller/provider/xavier/base/client.go
mansam: Likewise, `in` and `out` here both look like they are only ever a `base.Object`.

pkg/controller/provider/xavier/vsphere/adapter.go
```diff
@@ -0,0 +1,78 @@
...
+type Adapter struct {
+	// DB
+	libmodel.DB
+}
+
+//
+// Update the vSphere xavier body.
+func (r *Adapter) UpdateBody(object base.Object, m libmodel.Model) (err error) {
```
mansam, line 19: This returns an error, but its only caller ignores the error. Is that an oversight, or deliberate?

pkg/controller/provider/xavier/base/client.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+	client := http.Client{}
+	response, err := client.Do(request)
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+	status = response.StatusCode
+	content := []byte{}
+	if status == http.StatusOK {
```
mansam, line 269: This appears to be ignoring statuses other than 200. Is there going to be other error handling added later?

mansam (commented):
Looks good, just a small handful of questions.

pkg/controller/provider/xavier/base/client.go
jortel: Yes, the post() can be narrowed to an `*Object` for both in and out.

pkg/controller/provider/xavier/base/client.go
jortel: Not really.  Depending on the path, the `out` could be a `List`, string or anything.   The `out` is used as `in` in other places so even for `in` to be typed as `*Object` would make things difficult.

pkg/controller/provider/xavier/vsphere/adapter.go
jortel: Oversight.  Fixed.

pkg/controller/provider/xavier/base/client.go
jortel: No.  Anything besides a 200 is an error.

mansam (approved):

PR #79, opened 2020-11-14
Enable TLS for inventory

fdupont-redhat:
This pull request enabled TLS for the inventory container. If a secret name is passed as the `TLS_SECRET_NAME` environment variable, the TLS configuration is enabled and passed to the inventory container (see [konveyor/controller#29](https://github.com/konveyor/controller/pull/29)). The web client is also configured to use TLS and uses the openshift-signer CA certificate to validate the inventory service certificate.

Depends on https://github.com/konveyor/controller/pull/29.

### DIFF
--- a/pkg/controller/provider/web/base/client.go
+++ b/pkg/controller/provider/web/base/client.go
@@ -1,6 +1,8 @@
 package base
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -12,6 +14,8 @@ import (
 	liburl "net/url"
 	"reflect"
 )
+
+const ServiceCAFile = "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt"
 
 //
 // Application settings.
@@ -108,7 +112,18 @@ func (c *Client) get(path string, resource interface{}) (status int, err error) 
 		Header: header,
 		URL:    c.url(path),
 	}
-	client := http.Client{}
+	rootCAPool := x509.NewCertPool()
+	rootCA, err := ioutil.ReadFile(ServiceCAFile)
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+	rootCAPool.AppendCertsFromPEM(rootCA)
+	client := http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{RootCAs: rootCAPool},
+		},
+	}
 	response, err := client.Do(request)
 	if err != nil {
 		err = liberr.Wrap(err)
--- a/pkg/settings/inventory.go
+++ b/pkg/settings/inventory.go
@@ -14,6 +14,7 @@ const (
 	AuthOptional   = "AUTH_OPTIONAL"
 	Host           = "API_HOST"
 	Port           = "API_PORT"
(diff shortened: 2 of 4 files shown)


pkg/controller/provider/web/base/client.go
```diff
@@ -13,6 +15,8 @@ import (
 	"reflect"
 )
 
+const ServiceCAFile = "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt"
```
jortel, line 18: I don't think we want hard coded paths.
Recommend adding:
```
Inventory.TLS struct {
      ...
      // CA certificate path.
      CA string
}
```

pkg/settings/inventory.go
```diff
@@ -14,6 +14,7 @@ const (
 	AuthOptional   = "AUTH_OPTIONAL"
 	Host           = "API_HOST"
 	Port           = "API_PORT"
+	TLSSecretName  = "TLS_SECRET_NAME"
```
jortel, line 17: Few things.  The name of the secret would not be sufficient without the namespace.
Instead of the name of the secret, let's have the configmap/operator set the paths explicitly.  This will eliminate the need for hard coded paths and make it much easier to test and run in development environment.
API_TLS_KEY
API_TLS_CERTIFICATE
API TLS_CA

To be set by the operator.  This will k

pkg/settings/inventory.go
```diff
@@ -73,6 +83,14 @@ func (r *Inventory) Load() error {
 	} else {
 		r.Port = 8080
 	}
+	// TLS
+	if s, found := os.LookupEnv(TLSSecretName); found {
```
jortel, line 87: See comment above re: explicit paths.

pkg/controller/provider/web/base/client.go
fdupont-redhat: This is a hard coded path in OpenShift.

pkg/settings/inventory.go
fdupont-redhat: To be mounted, the secret must be in the same namespace as the pod, so the secret name is sufficient.
And the path is already set by the operator and follows the convention `/var/run/secrets/<secret_name>`.
Then, inside the secret, the file names are always `tls.crt` and `tls.key`.

jortel: I still think explicit paths are better than integration _by-convention_.  This way the operator can change it's convention without breaking the controller.  This decoupling is also useful for developement/testing.

pkg/settings/inventory.go
```diff
@@ -73,6 +84,16 @@ func (r *Inventory) Load() error {
...
 		r.Port = 8080
 	}
+	// TLS
+	if cert, found := os.LookupEnv(TLSCertificate); found {
+		if key, found := os.LookupEnv(TLSKey); found {
+			r.TLS.Certificate = cert
+			r.TLS.Key = key
+			r.TLS.Enabled = true
+		}
+	} else {
```
jortel, line 94: The `else` fine but not needed.  It will be initialized as `false`.

jortel (approved):
LGTM
This PR needs common controller [pull/29](https://github.com/konveyor/controller/pull/29) vendor'd in before merge.

PR #81, opened 2020-11-17
Fix TLS settings for inventory

fdupont-redhat:
Follow-up for https://github.com/konveyor/virt-controller/pull/79 to fix the TLS settings expansion.


jortel (approved):
LGTM

PR #85, opened 2020-11-18
Use internal mapping in VirtualMachineImport CR

### LINKED ISSUE #84: Embed mapping in VirtualMachineImport CR
In the current implementation, migrating a virtual machine generates two custom resources:

- ResourceMapping for the networks and storages of the virtual machine
- VirtualMachineImport to trigger the import by VMIO. This CR references the ResourceMapping CR.

When the migration is finished, either successful or failed, the custom resources are not deleted (expected). Over time, and even more when some migrations failed, the number of custom resources grows, as well as the required API calls to have a holistic view of the migrations

The ResourceMapping custom resources belonging to only one VirtualMachineImport, it would reduce the number of custom resources by a factor of two to store the network et storage mappings in the VirtualMachineImport custom resource.

RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1898939

fdupont-redhat:
This pull request changes the VirtualMachineImport build to:

- Snapshot the plan map in `plan.kubevirt.buildImport` and passing it to `plan.builder.vsphere.builder.Import`.
- Modify `plan.builder.vsphere.builder.Mapping` to embed the `VMImport.Spec.Source.Vmware.Mappings`.
- Modify `plan.builder.vsphere.builder.Mapping` to write the mappings in `VMImport.Spec.Source.Vmware.Mappings`.

Closes #84

### DIFF
--- a/pkg/controller/plan/builder/doc.go
+++ b/pkg/controller/plan/builder/doc.go
@@ -19,9 +19,9 @@ type Builder interface {
 	// Build secret.
 	Secret(vmID string, in, object *core.Secret) (err error)
 	// Build VMIO resource mapping.
-	Mapping(mp *plan.Map, object *vmio.ResourceMapping) error
+	Mapping(mp *plan.Map, object *vmio.VirtualMachineImportSpec) error
 	// Build VMIO import spec.
-	Import(vmID string, object *vmio.VirtualMachineImportSpec) error
+	Import(vmID string, mp *plan.Map, object *vmio.VirtualMachineImportSpec) error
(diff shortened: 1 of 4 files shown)


jortel (approved):
Well done.
Should be able to remove: Builder.[nameForMapping()](https://github.com/konveyor/virt-controller/blob/66b309a6252635014de97631d21cb2a59602e8b0/pkg/controller/plan/kubevirt.go#L237). Right?

fdupont-redhat:
@jortel good catch, I removed it in a new commit.

PR #86, opened 2020-11-19
Fix inventory client when TLS is enabled

fdupont-redhat:
With #79 merged, the inventory client doesn't find the right environment variables on the `main` container, so it is still trying to connect to http://localhost:8080. This pull request adds a `API_TLS_ENABLED` environment variable that affects the scheme.

The `API_HOST` and `API_PORT` environment variables can also be used to affect the rest of the base URL. This is done in [konveyor/virt-operator#44](https://github.com/konveyor/virt-operator/pull/44) to ensure the hostname matches the certificate subject.


jortel (approved):
LGTM

PR #87, opened 2020-11-19
Add type to VMIO network mapping

fdupont-redhat:
ResourceMapping sections of VirtualMachineImport CRs doesn't have the NetworkMapping.Type (pod/multus).
When only one network is mapped, this works because VMIO default is pod. However, it fails when there's more than one network as they all default to `pod` and a VM can only be connected once to the pod network.

This pull request propagates the type to the VMIO ResourceMapping.


jortel (approved):
LGTM

PR #89, opened 2020-11-19
Fix TSL when not enabled skip loading CA certificate.

jortel:
Fails with _file-not-found_ unless a certificate exists at /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt even with TLS  not enabled.  Also, cache the transport to avoid loading it for each Get() request.  When checking VMs, this could be thousands of times.  Although could have copied my test CA to the pod location (/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt) when running/testing locally, it was a bit of a pain.  The CA setting provides completeness and more flexibility.

Reduced _some_ complexity (deep nesting) in Settings.Inventory.Load() re: TLS settings.

### DIFF
--- a/pkg/controller/provider/web/base/client.go
+++ b/pkg/controller/provider/web/base/client.go
@@ -15,8 +15,6 @@ import (
 	"reflect"
 )
 
-const ServiceCAFile = "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt"
-
 //
 // Application settings.
 var Settings = &settings.Settings
@@ -44,6 +42,8 @@ type Client struct {
 	Host string
 	// Parameters
 	Params Params
+	// http Transport.
+	transport http.RoundTripper
 }
 
 //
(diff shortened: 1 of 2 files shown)


pkg/controller/provider/web/base/client.go
```diff
@@ -44,6 +42,8 @@ type Client struct {
 	Host string
 	// Parameters
 	Params Params
+	// http Transport.
+	transport http.RoundTripper
```
fdupont-redhat, line 46: Shouldn't it be `Transport` with an uppercase `T`?

jortel: The uppercase `Transport` would result in the filed being _exported_ (public outside the package).  The purpose of the field is to cache the transport internally so that it is only created once not so that it can be supplied on construction.  However, if you think there is a valid use case for the _Transport_ to be supplied, I don't see any potential danger in supporting it.

fdupont-redhat (approved):
LGTM

PR #95, opened 2020-12-02
Pass ESX host oriented MOID when migrating using the ESX host.

### LINKED ISSUE #70: Allow using names instead of ids in plans and mappings
When a user creates a CR for a plan or a mapping, the VMware networks, storages, VMs... are referenced by id. The UI displays names, then translates them to ids when creating the CRs. However, for a user that would like to interact only with the API, finding these IDs is not trivial, as they are not displayed in the vCenter UI.

Then comes the uniqueness issue of names. so it would require providing additional information when using names:

- VM name is unique withing the folder where it is located. So, the unique key is the absolute path of the VM.
- Datastore scope the datacenter. So, the unique key is the combination of datacenter name and datastore name.
- Standard vSphere Network scope is the host. Given that the VM belong to only one host, the unique key is the network name.
- Distributed vSwitch scope is the datacenter. So, the unique key is the combination of datacenter name and network name. This way, if the datacenter is provided we can expect the network to be a Distributed vSwitch.

### LINKED ISSUE #88: Translate mapped network/storage IDs to host be oriented.
When a `Host` is used for the migration (instead of vSphere API), the controller needs to translate mapped network/storage IDs to be host oriented when building the VM Import CR.

Likely to be easier to pass the `Name` instead of the `ID`.

jortel:
To support migration from ESX hosts, the mapped network & datastore ESX host oriented MOID needs to be passed instead of the vCenter MOID.

The ESX host lookup requires some of the same things needed to support VM, Network, Datastore referenced by `Name` as requested in issues/#70.  This PR will add support for either ID **and name** references in resources for:
- VM in plan CR.
- Host in Host CR
- Network in Plan and NetworkMap CRs.
- Datastore in Plan and StorageMap CRs.

Removed `Mapping()` from the `Builder` interface and changed to non-exported.  No reason for it to be _public_.
Moved a few methods to follow convention of _public_ (exported) methods at the top of the file with non-exported methods at the end.

After working with the API client, it seemed better for the `ProviderClient` to return errors instead of http status.  The following new error types added:
- web.NotFoundErr
- web.RefNotUniqueErr

---

Change summary:
- web.Client updated to return errors instead of http status.
- web.Client.Find() added to _find_ refs.
- web handlers for host, network, datastore and vm - added filter() to do path matching for List().
- code changed to use inventory.Find() instead of inventory.Get().
- add validations for _ambiguous_ references.
- CRDs for: host, network, datastore and plan updated to support either ID or Name references.
- Builder: now responsible for building the hostMap.
- Builder only map network/datastore referenced by the VM when building the VMIO mapping.
- Builder translates IDs to host oriented when using ESX host.

---

TODO:

The network name will need to be qualified by _type_ because both standard  and DV port groups can have the same name.

---

closes #70 
closes #88


config/crds/forklift_v1alpha1_networkmap.yaml
```diff
@@ -53,11 +53,15 @@ spec:
...
+                          ID.'
+                        type: string
+                      name:
+                        description: 'An object Name. vsphere:   A qualified name.'
+                        type: string
+                      type:
+                        description: Kind.
                         type: string
-                    required:
-                    - id
```
jortel, line 60: required `id`  replaced with optional (one of):
- id
- name
- type
validated by the controller.

pkg/apis/forklift/v1alpha1/host.go
```diff
@@ -18,19 +18,17 @@ package v1alpha1
...
 	libcnd "github.com/konveyor/controller/pkg/condition"
+	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1alpha1/ref"
 	core "k8s.io/api/core/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 //
 // HostSpec defines the desired state of Host
 type HostSpec struct {
+	ref.Ref `json:",inline"`
```
jortel, line 29: ref.Ref is shared reference with optional (id, name, type).

pkg/apis/forklift/v1alpha1/ref/ref.go
```diff
@@ -0,0 +1,38 @@
+package ref
+
+//
+// Source reference.
+// Either the ID or Name must be specified.
+type Ref struct {
```
jortel, line 6: Renamed, moved: mapping.SourceObject now used for many other things.

pkg/controller/host/validation.go
```diff
@@ -133,20 +136,32 @@ func (r *Reconciler) validateID(host *api.Host) error {
 	default:
 		return nil
 	}
-	status, err := inventory.Get(resource, host.Spec.ID)
+	err = inventory.Find(resource, ref)
```
jortel, line 139: Uses Client.Find() by ref instead of ID-based Get().

pkg/controller/host/validation.go
```diff
@@ -133,20 +136,32 @@ func (r *Reconciler) validateID(host *api.Host) error {
...
+					Status:   True,
+					Reason:   Ambiguous,
+					Category: Critical,
+					Message:  "Host reference is ambiguous.",
+				})
+			return nil
+		}
 		return liberr.Wrap(err)
 	}
-	if status != http.StatusOK {
```
jortel, line 140: The web.Client methods updated to return errors for not-found and ambiguous find result.

pkg/controller/plan/builder/doc.go
```diff
@@ -16,31 +17,35 @@ import (
 // Builds/updates objects as needed with provider
 // specific constructs.
 type Builder interface {
+	// Resolve the VM ref.
+	Resolve(vmRef *ref.Ref) error
 	// Build secret.
-	Secret(vmID string, in, object *core.Secret) (err error)
-	// Build VMIO resource mapping.
-	Mapping(mp *plan.Map, object *vmio.VirtualMachineImportSpec) error
+	Secret(vmRef ref.Ref, in, object *core.Secret) error
```
jortel, line 23: Need to pass the ref.Ref instead of ID.

pkg/controller/plan/builder/doc.go
```diff
@@ -16,31 +17,35 @@ import (
...
+	Tasks(vmRef ref.Ref) ([]*plan.Task, error)
 }
 
 //
 // Builder factory.
 func New(
 	client client.Client,
 	inventory web.Client,
-	provider *api.Provider,
-	hostMap map[string]*api.Host) (builder Builder, err error) {
```
jortel, line 35: The hostMap needs to be created by the builder instead of passed in.

pkg/controller/plan/builder/doc.go
```diff
@@ -16,31 +17,35 @@ import (
...
 	switch provider.Type() {
 	case api.VSphere:
-		builder = &vsphere.Builder{
+		b := &vsphere.Builder{
 			Client:    client,
 			Inventory: inventory,
 			Provider:  provider,
-			HostMap:   hostMap,
+		}
+		bErr := b.Load()
```
jortel, line 44: load the hostMap.

pkg/controller/plan/kubevirt.go
```diff
@@ -183,10 +189,15 @@ func (r *KubeVirt) buildImport(vm *plan.VMStatus) (object *vmio.VirtualMachineIm
 		err = liberr.Wrap(err)
 		return
 	}
+	err = r.Builder.Resolve(&vm.Ref)
```
jortel, line 192: The Builder.Resolve() will do an inventory lookup and ensure all of the fields (both ID and Name) are populated.  This is needed for naming created CRs and other actions to follow.

pkg/controller/plan/migration.go
```diff
@@ -104,8 +104,6 @@ type Migration struct {
 	kubevirt KubeVirt
 	// VM import CRs.
 	importMap ImportMap
-	// Host map.
-	hostMap map[string]*api.Host
```
jortel, line 108: Created by the builder now.

pkg/controller/plan/migration.go
```diff
@@ -583,32 +575,6 @@ func (r *Migration) updatePipeline(vm *plan.VMStatus, imp *VmImport) {
 	}
 }
 
-//
-// Build the host map (as needed).
-func (r *Migration) buildHostMap() (err error) {
```
jortel, line 588: Created by the builder now.

pkg/controller/plan/validation.go
```diff
@@ -112,35 +115,62 @@ func (r *Reconciler) validateVM(provider *api.Provider, plan *api.Plan) error {
...
+		Category: Critical,
+		Message:  "VM reference is ambiguous.",
+		Items:    []string{},
+	}
+	setOf := map[string]bool{}
+	hash := func(object interface{}) string {
+		h := sha256.New()
+		h.Write([]byte(fmt.Sprintf("%v", resource)))
+		return string(h.Sum(nil))
+	}
```
jortel, line 142: web.Client.Find() can match more than 1 VM depending on how much of the absolute path is specified as part of the name.

Also, for uniqueness, we need to handle the case whereby 2 VMs - one specified by name and the other by ID is actually the same VM.  Duplicate detection becomes a little more complex.

pkg/controller/provider/web/base/client.go
```diff
@@ -23,15 +24,59 @@ var Settings = &settings.Settings
...
 //
 // Resolves resources to API paths.
 type Resolver interface {
 	// Find the API path for the specified resource.
 	Path(resource interface{}, id string) (string, error)
 }
 
+//
+// Resource Finder.
+type Finder interface {
```
jortel, line 44: New `Finder` interface implemented by each provider _type_ support package.

pkg/controller/provider/web/base/client.go
```diff
@@ -23,15 +24,59 @@ var Settings = &settings.Settings
...
+	//   ProviderNotReadyErr
+	//   NotFoundErr
+	List(list interface{}, param ...Param) error
+	// Get a resource by ref.
+	// Returns:
+	//   ProviderNotSupportedErr
+	//   ProviderNotReadyErr
+	//   NotFoundErr
+	//   RefNotUniqueErr
+	Find(resource interface{}, ref Ref) error
```
jortel, line 77: For convenience, added the `Find()` method to support finding resources using the `Ref` (id, name, type).

pkg/controller/provider/web/base/client.go
```diff
@@ -41,7 +86,7 @@ type Param struct {
 
 //
 // REST API client.
-type Client struct {
+type RestClient struct {
```
jortel, line 89: Renamed for clarity.

pkg/controller/provider/web/client.go
```diff
@@ -11,23 +11,20 @@ import (
...
-	Get(resource interface{}, id string) (int, error)
-	// List a collection.
-	// The `list` must be a pointer to a slice of resource object.
-	// Returns: The HTTP code and error.
-	List(list interface{}) (int, error)
-}
+// Interfaces.
+type Client = base.Client
+type Finder = base.Finder
+type Param = base.Param
```
jortel, line 27: Aggregated export of the base types.  Can be imported and used as `web.Client` etc.

pkg/controller/provider/web/client.go
```diff
@@ -57,52 +56,92 @@ func NewClient(provider *api.Provider) (client Client, err error) {
...
+	finder Finder
 	// Ready.
 	found bool
 }
 
 //
 // Get a resource.
-// Raises ProviderNotReadyErr on 404 and 206
-// when the provider is not yet in the inventory.
-func (r *ProviderClient) Get(resource interface{}, id string) (status int, err error) {
```
jortel, line 71: The `ProviderClient` no longer returns status but specific error types instead to improve usability at this layer.

pkg/controller/provider/web/ocp/client.go
```diff
@@ -22,7 +22,7 @@ func (r *Resolver) Path(object interface{}, id string) (path string, err error)
 	ns = strings.TrimSuffix(ns, "/")
 	switch object.(type) {
 	case *Provider:
-		if id == "" { // list
+		if id == "/" { // list
```
jortel, line 25: This was just broken.

pkg/controller/provider/web/ocp/client.go
```diff
@@ -69,3 +69,15 @@ func (r *Resolver) Path(object interface{}, id string) (path string, err error)
...
+
+//
+// Resource finder.
+type Finder struct {
+	base.Client
+}
+
+//
+// Find a resource by ref.
+func (r *Finder) ByRef(client base.Client, resource interface{}, ref base.Ref) (err error) {
```
jortel, line 81: Added just to comply with the interface.  I don't anticipate this will ever be implemented.

pkg/controller/provider/web/vsphere/base.go
```diff
@@ -22,10 +23,52 @@ type Handler struct {
 // Build list predicate.
 func (h Handler) Predicate(ctx *gin.Context) (p libmodel.Predicate) {
 	q := ctx.Request.URL.Query()
-	value := q.Get(NameParam)
-	if len(value) > 0 {
-		p = libmodel.Eq(NameParam, value)
+	name := q.Get(NameParam)
```
jortel, line 26: Needed to support path qualified names.

pkg/controller/provider/web/vsphere/client.go
```diff
@@ -82,3 +89,131 @@ func (r *Resolver) Path(resource interface{}, id string) (path string, err error
...
 }
+
+//
+// Resource finder.
+type Finder struct {
+}
+
+//
+// Find a resource by ref.
+func (r *Finder) ByRef(client base.Client, resource interface{}, ref base.Ref) (err error) {
```
jortel, line 100: This is the _meat_ of the find-by-ref capability.
Performs a `List()` and if the query matches multiple resources (list contains more than one) raises and `ambiguous` error.  Else, updates the object passed in with list[0].

pkg/controller/provider/web/vsphere/datastore.go
```diff
@@ -55,9 +57,23 @@ func (h DatastoreHandler) List(ctx *gin.Context) {
 		return
 	}
 	content := []interface{}{}
+	err = h.filter(ctx, &list)
```
jortel, line 60: Query by name.  Then, filter by path match.

pkg/controller/plan/builder/vsphere/builder.go
```diff
@@ -82,28 +104,124 @@ func (r *Builder) Secret(vmID string, in, object *core.Secret) (err error) {
...
+				},
+			})
 	}
 
 	return
 }
 
+//
+// Load (as needed).
+func (r *Builder) Load() (err error) {
```
jortel, line 168: Now responsible for building the hostMap no longer passed in.

pkg/controller/plan/builder/vsphere/builder.go
```diff
@@ -148,38 +260,80 @@ func (r *Builder) host(hostID string) (host *vsphere.Host, err error) {
...
+			return
+		}
+		needed := false
+		for _, net := range vm.Networks {
+			if net.ID == network.ID {
+				needed = true
+				break
+			}
+		}
+		if !needed {
```
jortel, line 281: Determine if the VM references the network and conditionally included in the VMIO map.  This is more efficient and necessary because a network/datastore for another VM may not be known to this ESX host.

pkg/controller/plan/builder/vsphere/builder.go
```diff
@@ -148,38 +260,80 @@ func (r *Builder) host(hostID string) (host *vsphere.Host, err error) {
...
+			continue
+		}
+		id, pErr := r.networkID(vm, network)
```
jortel, line 284: Translated to ESX host ID as needed here.

pkg/controller/provider/web/client.go
```diff
@@ -57,52 +56,92 @@ func NewClient(provider *api.Provider) (client Client, err error) {
...
 	return
 }
 
 //
 // List a resource collection.
-// Raises ProviderNotReadyErr on 404 and 206
-// when the provider is not yet in the inventory.
-func (r *ProviderClient) List(resource interface{}) (status int, err error) {
-	err = r.Find()
+// Raises:
```
mansam, line 102: It now also raises a `NotFoundErr`

mansam (approved):
Looks good, just a couple minor things

pkg/apis/forklift/v1alpha1/ref/ref.go
```diff
@@ -0,0 +1,38 @@
...
+	// The object ID.
+	// vsphere:
+	//   The managed object ID.
+	ID string `json:"id,omitempty"`
+	// An object Name.
+	// vsphere:
+	//   A qualified name.
+	Name string `json:"name,omitempty"`
+	// Kind.
+	Kind string `json:"type,omitempty"`
```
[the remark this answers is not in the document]
jortel: Fixed.

PR #97, opened 2020-12-02
Optional host secret not fully supported.

jortel:
The secret referenced in the `Host` CR is optional.  The CRD specified this but not completely supported in the controller.  Needed to relax the host validation and handle appropriately when building the VMIO secret.

### DIFF
--- a/pkg/controller/host/validation.go
+++ b/pkg/controller/host/validation.go
@@ -174,18 +174,16 @@ func (r *Reconciler) validateIp(host *api.Host) error {
 //   2. The secret exists.
 //   3. the content of the secret is valid.
 func (r *Reconciler) validateSecret(host *api.Host) error {
-	// NotSet
+	ref := host.Spec.Secret
+	if !libref.RefSet(ref) {
+		return nil
+	}
 	newCnd := libcnd.Condition{
 		Type:     SecretNotValid,
 		Status:   True,
-		Reason:   NotSet,
+		Reason:   NotFound,
 		Category: Critical,
 		Message:  "The `secret` is not valid.",
-	}
-	ref := host.Spec.Secret
-	if !libref.RefSet(&ref) {
-		host.Status.SetCondition(newCnd)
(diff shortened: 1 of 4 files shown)


pkg/apis/virt/v1alpha1/host.go
```diff
@@ -36,7 +36,7 @@ type HostSpec struct {
 	// Certificate SHA-1 fingerprint, called thumbprint by VMware.
 	Thumbprint string `json:"thumbprint,omitempty"`
 	// Credentials.
-	Secret core.ObjectReference `json:"secret,omitempty"`
+	Secret *core.ObjectReference `json:"secret,omitempty"`
```
jortel, line 39: Making a pointer makes the yaml easier in that the property can be omitted all together instead of:
```
secret: {}
```

pkg/controller/host/validation.go
```diff
@@ -174,19 +174,17 @@ func (r *Reconciler) validateIp(host *api.Host) error {
 //   2. The secret exists.
 //   3. the content of the secret is valid.
 func (r *Reconciler) validateSecret(host *api.Host) error {
-	// NotSet
+	ref := host.Spec.Secret
+	if !libref.RefSet(ref) {
```
jortel, line 178: Skip further validation.

pkg/controller/plan/builder/vsphere/builder.go
```diff
@@ -47,18 +48,20 @@ func (r *Builder) Secret(vmID string, in, object *core.Secret) (err error) {
 			Path:   vim25.Path,
 		}
 		url = hostURL.String()
-		hostSecret, nErr := r.hostSecret(host)
-		if nErr != nil {
-			err = liberr.Wrap(nErr)
-			return
+		if libref.RefSet(host.Spec.Secret) {
```
jortel, line 51: Skip here.

mansam (approved):

PR #99, opened 2020-12-02
Replace Travis by Github Actions

fdupont-redhat:
In Konveyor org, we have hit the Travis quota, so we transition to Github Actions.


jortel (approved):
LGTM

PR #100, opened 2020-12-03
Add Golang setup step

fdupont-redhat:
When enabling Github Actions, we need to setup the Golang environment before running the tests.


fdupont-redhat:
Just fixing CI. Not waiting for review.

PR #112, opened 2020-12-10
Improved migration error handling; Typed errors using errors.As().

jortel:
Currently all errors from building and creating the VMIO CR are bubbled up to main Reconcile(), traced and the reconcile re-queued.  Instead, errors associated with building/creating the CR should be captured as an `Error` on the VM pipeline structure itself.  As a result, the individual import would be failed.

To accomplish this, we really needed to include more contextual information in the errors being bubbled up.  To do this, we also needed raise/check errors based on error type using errors.As() instead of errors.Is() which does a direct comparison.  This is because to include the contextual information, we need to create a _new_ instance of the error.

### DIFF
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -2,6 +2,7 @@ package plan
 
 import (
 	"context"
+	"errors"
 	libcnd "github.com/konveyor/controller/pkg/condition"
 	liberr "github.com/konveyor/controller/pkg/error"
 	libitr "github.com/konveyor/controller/pkg/itinerary"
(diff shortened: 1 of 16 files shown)


pkg/controller/plan/migration.go
```diff
@@ -152,13 +153,25 @@ func (r Migration) Run() (reQ time.Duration, err error) {
...
-				err = liberr.Wrap(err)
-				return
+				if !errors.As(err, &web.ProviderNotReadyError{}) {
+					vm.AddError(err.Error())
+					log.Trace(err)
+					err = nil
+					break
+				} else {
+					return
+				}
```
jortel, line 174: The main fix here ^^.

mansam (approved):

Issue #77, opened 2020-11-12
Add support for access and volume mode in storage mapping


fdupont-redhat:
When a storage mapping is created, the only information we have is the destination storage class name. However, this is not sufficient, because some access modes and volume modes are not support by Kubevirt, e.g. Ceph + Filesystem or Cinder + RWX. So, we shouldn't allow unsupported access and volume modes.

The idea is to implement a storage support matrix that would be used to validate access and volume modes for the destination storage classes, as well as provide _intelligent_ default values for each provisioner type. The support matrix would be stored in a Config Map managed by the operator and mounted in the controller and UI pods.

The access modes are documented in [Kubernetes > Persistent Volumes > Access Modes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes).
The volumes modes are documented in [Kubernetes > Persistent Volumes > Volume Modes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#volume-modes).
The list of provisioners that support block mode is available in [Kubernetes > Persistent Volumes > Raw Block Volume Support](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#raw-block-volume-support).

A possible format for the support matrix is:

```yaml
---
- kubevirt_version: v2.5.0
  supported_storages:
  - provisioner_name: kubernetes.io/aws-ebs
    access_modes:
    - ReadWriteOnce
    volume_modes:
    - Block
    - Filesystem
  - provisioner_name: kubernetes.io/rbd
    access_modes:
    - ReadWriteOnce
    - ReadOnlyMany
    volume_modes:
    - Block
    - Filesystem
  - provisioner_name: kubernetes.io/glusterfs
    access_modes:
    - ReadWriteMany
    - ReadWriteOnce
    - ReadOnlyMany
    volume_modes:
    - Filesystem
```

jortel:
Recommend:
```
---
- kubevirtVersion: v2.5.0
  provisioners:
  - name: kubernetes.io/aws-ebs
    accessModes:
    - ReadWriteOnce
    volumeModes:
    - Block
    - Filesystem
 ```

mansam:
VMIO currently supports setting the volume mode on the Storage and Disk mappings, and I'm working on support for setting the access mode.

```go
// StorageResourceMappingItem defines the storage mapping of a single resource from the provider to kubevirt
// +k8s:openapi-gen=true
type StorageResourceMappingItem struct {
	Source Source           `json:"source"`
[...]

PR #115, opened 2020-12-10
Add provisioner CRD.

jortel:
Add provisioner CRD.  We can follow up with _Reconciler_ to validate later.

Re: #77


fdupont-redhat (approved):
LGTM

PR #92, opened 2020-11-24
Transition to go modules for dep management

shawn-hurley:
* remove Gopkg.{lock/toml}
* Update Makefile to pull controller-gen correctly and use it
* Updated module names for kubevirt imports
* Updated client.List calls to conform to correct dynamic client interface.

fixes #82

PR #126, opened 2020-12-16
Transition to go modules for dep management

### LINKED ISSUE #82: Move from dep to go modules
With Go 1.14, modules are the standard for dependency management.

fdupont-redhat:
Since the creation of #92, the code has changed and the repository has been renamed, so rebasing seems over complex. This PR is doing the same changes, but from a more recent commit.

* remove Gopkg.{lock/toml}
* Update Makefile to pull controller-gen correctly and use it
* Updated module names for kubevirt imports
* Updated client.List calls to conform to correct dynamic client interface.

fixes #82


jortel (approved):
LGTM

PR #130, opened 2020-12-17
Set volumeMode only when specified or defaulted.

jortel:
Set `volumeMode` in the VMIO CR mapping only when specified or defaulted.


fdupont-redhat (approved):
LGTM

PR #132, opened 2020-12-17
Reconciler improvements.

jortel:
General reconciler improvements and standardization.

We don't want to trace provider-not-ready errors.  This add a lot of unnecessary noise in the logs.

### DIFF
--- a/pkg/controller/map/storage/controller.go
+++ b/pkg/controller/map/storage/controller.go
@@ -40,7 +40,7 @@ const (
 	// Controller name.
 	Name = "storage-map"
 	// Fast re-queue delay.
-	FastReQ = time.Millisecond * 100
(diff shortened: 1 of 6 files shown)


pkg/controller/map/storage/controller.go
```diff
@@ -105,27 +105,37 @@ type Reconciler struct {
...
+		if err != nil {
+			log.Trace(err)
+			err = nil
+		}
+	}()
+
 	// Fetch the CR.
 	mp := &api.StorageMap{}
 	err = r.Get(context.TODO(), request.NamespacedName, mp)
+	err = r.Get(context.TODO(), request.NamespacedName, mp)
```
mansam, line 129: looks like this line was accidentally duplicated

mansam (approved):

PR #140, opened 2021-01-13
Fix host connection test.

jortel:
The connection test was still using the `thumbprint` in the secret instead of the thumbprint for the host in the inventory.


pkg/controller/plan/builder/vsphere/host.go
```diff
@@ -22,8 +21,6 @@ type EsxHost struct {
 	URL string
 	// Host secret.
 	Secret *core.Secret
-	// Inventory client.
-	Inventory web.Client
```
jortel, line 26: Unused.

mansam (approved):

PR #143, opened 2021-01-19
Add a new composite VM resource to the provider API

jortel:
Add a _new_ composite resource (endpoint) to the provider inventory/API.  Borrowing the term from _vSphere_, the new resource is called a _"workload"_ and contains all fields as a `VM` but with the `host` field redefined to be the de-referenced `Host` resource.  The (de-referenced) `Host` resource has all of the same properties as a regular `Host` but with an additional `cluster` property.  The `Cluster` has all the same fields a a regular `Cluster` but with an additional `datacenter` field which contains the `Datacenter`.

This composite resource is generally useful.  However, an immediate use case is: for the policy-agent to evaluate rules against the VM and its host and cluster.

To support this, the _tree_ functionality in the model and web layers was expanded to support:
- Build a tree from the leaf up (ancestry).
- Make tree navigation more configurable.  As a result, the hard coded navigation was replaced with _caller_ supplied `Navigator` objects which guides the tree builder as it traverses up/down the tree.
- Removed unused _flattening_ functionality from the `Tree` as it has not proven to be valuable.

### DIFF
--- a/pkg/controller/provider/model/vsphere/model.go
+++ b/pkg/controller/provider/model/vsphere/model.go
@@ -1,7 +1,6 @@
 package vsphere
 
 import (
-	liberr "github.com/konveyor/controller/pkg/error"
 	libmodel "github.com/konveyor/controller/pkg/inventory/model"
 	"strings"
 )
@@ -166,6 +165,37 @@ type Ref struct {
 	Kind string `json:"kind"`
 	// The ID of object referenced.
 	ID string `json:"id"`
+}
+
+//
+// Get referenced model.
+func (r *Ref) Get(db libmodel.DB) (model Model, err error) {
+	base := Base{
+		ID: r.ID,
+	}
+	switch r.Kind {
+	case FolderKind:
+		model = &Folder{Base: base}
+	case DatacenterKind:
+		model = &Datacenter{Base: base}
+	case ClusterKind:
+		model = &Cluster{Base: base}
+	case HostKind:
+		model = &Host{Base: base}
+	case VmKind:
+		model = &VM{Base: base}
+	case NetKind:
+		model = &Network{Base: base}
+	case DsKind:
+		model = &Datastore{Base: base}
+	default:
+		err = InvalidRefError{*r}
+	}
+	if model != nil {
+		err = db.Get(model)
+	}
+
+	return
 }
 
 type About struct {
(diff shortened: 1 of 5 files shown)


pkg/controller/provider/model/vsphere/model.go
```diff
@@ -168,6 +167,37 @@ type Ref struct {
 	ID string `json:"id"`
 }
 
+//
+// Get referenced model.
+func (r *Ref) Get(db libmodel.DB) (model Model, err error) {
```
jortel, line 172: Moved from _tree.go_.

pkg/controller/provider/model/vsphere/model.go
```diff
@@ -337,32 +367,6 @@ func (m *VM) Analyzed() bool {
 	return m.RevisionAnalyzed == m.Revision
 }
 
-//
-// Find associated cluster.
-// returns err = libmodel.NotFound when cannot be resolved.
-func (m *VM) Cluster(db libmodel.DB) (matched *Cluster, err error) {
```
jortel, line 343: Unused.

pkg/controller/provider/model/vsphere/tree.go
```diff
@@ -45,25 +44,18 @@ func (r InvalidKindError) Error() string {
 type Tree struct {
 	// DB connection.
 	DB libmodel.DB
-	// Tree root.
-	Root Model
-	// Leaf kind.
-	Leaf string
-	// Flatten the tree (root & leafs).
-	Flatten bool
```
jortel, line 53: Flattening not proven to be useful.

pkg/controller/provider/model/vsphere/tree.go
```diff
@@ -45,25 +44,18 @@ func (r InvalidKindError) Error() string {
 type Tree struct {
 	// DB connection.
 	DB libmodel.DB
-	// Tree root.
-	Root Model
-	// Leaf kind.
-	Leaf string
```
jortel, line 51: Now that tree traversal is both down and up, the Root and Leaf only pertain to the `Build()`.

pkg/controller/provider/model/vsphere/tree.go
```diff
@@ -45,25 +44,18 @@ func (r InvalidKindError) Error() string {
...
+func (r *Tree) Build(root Model, navigator BranchNavigator) (*TreeNode, error) {
+	node := &TreeNode{
+		Kind:  libref.ToKind(root),
+		Model: root,
 	}
-	node := root
+	treeRoot := node
 	depth := 0
 	var walk func(Model, bool) error
 	walk = func(model Model, asChild bool) error {
```
jortel, line 61: This method mainly simplified to delegate traversal path to the `navigator`.

pkg/controller/provider/model/vsphere/tree.go
```diff
@@ -74,181 +66,67 @@ func (r *Tree) Build() (*TreeNode, error) {
...
 		return nil, liberr.Wrap(err)
 	}
 
-	return root, nil
+	return treeRoot, nil
 }
 
 //
-// Get referenced model.
-func (r *Tree) getRef(ref Ref) (model Model, err error) {
```
jortel, line 225: The `Get` functionality moved to the `Ref` object itself.  More generally useful.

pkg/controller/provider/model/vsphere/tree.go
```diff
@@ -74,181 +66,67 @@ func (r *Tree) Build() (*TreeNode, error) {
...
-	base := Base{
-		ID: ref.ID,
+// Build the (ancestry) tree.
+func (r *Tree) Ancestry(leaf Model, navigator ParentNavigator) (*TreeNode, error) {
```
jortel, line 105: Added so support building an _ancestry_ tree from the leaf.

mansam (approved):

PR #145, opened 2021-01-20
Plan snapshot redesigned.

jortel:
Revisited the _snapshot_ approach to capture information including generation of the plan and related resources instead of stashing the resources in the migration annotations.

Add migration _history_ in the plan status. Each entry contains a snapshot of the plan and related resources (references):
- conditions (Executing, Succeeded, Failed, Canceled)
- plan (ref)
- provider
  - source (ref)
  - destination (ref)
- migration.
Each snapshot `ref` contains:
- namespace
- name
- uid
- generation

The plan controller will use the history to detect that one (or more) of the resources has been modified while the migration.  When detected, the migration will be marked as canceled with reason=modified the migration.
Further, when a migration is deleted (while running), it will be marked as canceled with reason=deleted  in the history.
The migration controller will reflect the status (conditions) for it's snapshot in the history.

The history provides a straight forward and unambiguous method of tracking migration status that is not vulnerable to plan/migration controller reconcile race conditions and timing.  It also provides a good audit trail.

*This should not impact the UI.
**Requires CRD update in the operator.


pkg/apis/forklift/v1alpha1/provider.go
```diff
@@ -42,15 +42,6 @@ const (
 	Token = "token"
 )
 
-//
-// Provider pair.
-type ProviderPair struct {
```
jortel, line 47: Removing this because it does not add value.  Better to just use provider.Pair instead.

pkg/controller/migration/controller.go
```diff
@@ -195,33 +194,45 @@ func (r *Reconciler) reflectPlan(plan *api.Plan, migration *api.Migration) {
...
+		return
+	}
+	if snapshot.HasCondition(Executing) {
+		migration.Status.MarkStarted()
+		migration.Status.VMs = plan.Status.Migration.VMs
 		migration.Status.SetCondition(libcnd.Condition{
 			Type:     Running,
 			Status:   True,
-			Category: Required,
+			Category: Advisory,
```
jortel, line 216: Not related but noticed the category is incorrect.

mansam (approved):

PR #154, opened 2021-02-09
Support selective VM cancelation

mansam:
* Add `cancel` array of `Refs` to the MigrationSpec.
* Add `conditions` to VMStatus
* Regenerate CRDs with new fields
* Refactor `FindVM` on the PlanSpec and MigrationStatus to take a Ref as a parameter.
* Refactor usages of `FindVM` to resolve the Ref before calling.
* Set `Failed` and `Succeeded` conditions on VMs as appropriate when a migration ends
* Set `Canceled` condition on all incomplete VMs when the snapshot fails to match.
* Set `Canceled` condition with a `UserRequested` reason on VMs that are listed in the Migration's `cancel` field.
* Refactor the migration runner's `Cancel` method to clean up VMs with a `Canceled` or `Failed` condition.


pkg/controller/plan/migration.go
```diff
@@ -110,6 +110,23 @@ func (r Migration) Run() (reQ time.Duration, err error) {
...
+				libcnd.Condition{
+					Type:     Canceled,
+					Status:   True,
+					Category: Advisory,
+					Reason:   UserRequested,
+					Message:  "The migration has been canceled.",
+					Durable:  true,
+				})
+			vm.Phase = Completed
+			log.Info("Migration [CANCELED]:", "vm", vm)
```
jortel, line 126: I think this will be logged over-and-over for every reconcile which seems undesirable, right?

pkg/controller/plan/migration.go
```diff
@@ -454,8 +490,25 @@ func (r *Migration) updateVM(vm *plan.VMStatus) (completed bool, failed bool, er
 		vm.MarkedCompleted()
```
jortel, line 490: Unrelated but will bet it should be `MarkComplete()`.

pkg/controller/plan/migration.go
mansam: `Cancel()` calls `vm.MarkCompleted()` after deleting the resources, which should cause that VM to be skipped on any subsequent reconcile.

jortel: Gotcha.

pkg/controller/plan/migration.go
```diff
@@ -272,8 +301,17 @@ func (r *Migration) begin() (err error) {
 	//
 	// Delete
 	for _, status := range r.Plan.Status.Migration.VMs {
+
```
jortel, line 304: Intentional blank lines?

jortel (approved):
LGTM

PR #159, opened 2021-02-11
Fix condition flow between plan and snapshot

mansam:
This fixes a few things I didn't get quite right when switching the way the conditions are reflected.

* Finish staging conditions on the snapshot before cloning them to the plan.
* Fix erroneous usage of `HasCondition` where `HasAnyCondition` was needed.
* Remove any conditions from the plan that aren't on the snapshot.
* Set the `Executing` condition on the active snapshot instead of the plan.


mansam:
Moved back to draft, looks like there's something else to fix.

jortel (approved):
LGTM

PR #161, opened 2021-02-11
Wrap error (only) at the edge.

jortel:
The original philosophy was the by wrapping errors at all levels would:
- Ensure errors got wrapped.
- Easier to code review because it's consistent.
After living with it for a while it seem apparent that neither are true.  Wrapping only where errors are returned by external packages is:
- reliable
- straight forward to code review
- reduces code noise.


mansam (approved):
LGTM

PR #164, opened 2021-02-15
Update the version of vm-import-operator to 0.3.0

mansam:
This updates the version of vm-import-operator we depend on to one which includes the new warm migration fields.


go.mod
```diff
@@ -7,15 +7,15 @@ require (
 	github.com/k8snetworkplumbingwg/network-attachment-definition-client v1.1.0
 	github.com/konveyor/controller v0.2.8
 	github.com/onsi/gomega v1.10.3
-	github.com/prometheus/client_golang v1.8.0
+	github.com/prometheus/client_golang v1.8.0 // indirect
```
jortel, line 10: @shawn-hurley Any idea why this would change from (assume direct) to `//indirect`?  I can understand why a dep would be _upgraded_ from `//indirect` to _direct_ when directly used by our code (as opposed to dep) but not sure I understand what seems like a _downgrade_.
Or, perhaps I don't understand what `//indirect` signifies.

jortel (approved):
LGTM

go.mod
shawn-hurley: I assume the answer here is that one of the imports from client-go or containerized-data-importer used to directly import these, but now it only indirectly through a different package import this. I think that is the rules

PR #166, opened 2021-02-16
Provider API: Add Cluster.Hosts and include DvSwitch in /networks endpoint.

jortel:
Add (missing) Cluster.Hosts and include DvSwitch in /networks model & endpoint.
Add `variant` to the Network model and REST resource.


mansam (approved):
LGTM

PR #169, opened 2021-02-18
Mapping included in plan by reference.

jortel:
Refactor network and storage mappings to be by-reference instead of embedded in the Plan.
This includes:
- Plan spec `Map` section updated.
- Plan validation ensure references to mapping CRs are:
  - set and valid.
  - have `Ready` condition.
  - resolved and added to `Plan.Referenced`
- NetworkMap and StorageMap CRs are reconciled before Plan CRs.
- Mapping CRs are included in `PlanContext`.
- Mapping CRs are included in snapshots (and snapshot matching).
- Network and Storage validations moved from the shared `validation` package to their respective controllers.


pkg/controller/map/storage/validation.go
```diff
@@ -33,21 +45,135 @@ const (
...
+	list := mp.Spec.Map
+	for i := range list {
+		ref := &list[i].Source
+		if ref.NotSet() {
+			mp.Status.SetCondition(libcnd.Condition{
+				Type:     DestinationStorageNotValid,
+				Status:   True,
+				Reason:   NotSet,
+				Category: Critical,
+				Message:  "Destination network: either `ID` or `Name` required.",
```
mansam, line 90: I think the condition here is supposed to be `SourceStorageNotValid`, and the message should be `"Source storage: ..."`

mansam (changes requested):

mansam:
LGTM other than a copy/paste error with one of the conditions.

pkg/controller/map/storage/validation.go
jortel: Fixed.

mansam (approved):

PR #170, opened 2021-02-23
Add host subnetMask; managementServerIp.

### LINKED ISSUE #120: Expand network attributes of the Host model
When a vSphere provider is added, the UI leverages the inventory to list the hosts and allow selecting a migration network.
By default, the migration traffic will go over the management network, but the UI cannot display that information. However, the vSphere API exposes the IP address of the ESXi host used for management:

`MOR:HostSystem.summary.managementServerIp`.

If the inventory contained that attribute, the UI could then identify the default network used for migration, by doing some IP calculation based on the IP addresses and netmasks of the pNICs. The netmask is available as:

`MOR:HostSystem.configManager.networkSystem.networkInfo.vnic[*].spec.ip.subnetMask`

jortel:
Add host subnetMask; managementServerIp.

closes #120

### DIFF
--- a/pkg/controller/provider/model/vsphere/model.go
+++ b/pkg/controller/provider/model/vsphere/model.go
@@ -231,21 +231,22 @@ type Cluster struct {
 
 type Host struct {
 	Base
-	InMaintenanceMode bool        `sql:""`
-	Thumbprint        string      `sql:""`
-	CpuSockets        int16       `sql:""`
-	CpuCores          int16       `sql:""`
-	ProductName       string      `sql:""`
-	ProductVersion    string      `sql:""`
-	Network           HostNetwork `sql:""`
-	Networks          []Ref       `sql:""`
-	Datastores        []Ref       `sql:""`
-	Vms               []Ref       `sql:""`
+	InMaintenanceMode  bool        `sql:""`
+	ManagementServerIp string      `sql:""`
+	Thumbprint         string      `sql:""`
+	CpuSockets         int16       `sql:""`
+	CpuCores           int16       `sql:""`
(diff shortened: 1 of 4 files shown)


pkg/controller/provider/model/vsphere/model.go
```diff
@@ -231,21 +231,22 @@ type Cluster struct {
...
+	Networks           []Ref       `sql:""`
+	Datastores         []Ref       `sql:""`
+	Vms                []Ref       `sql:""`
 }
 
 type HostNetwork struct {
-	PNICs      []PNIC      `json:"vNICs"`
-	VNICs      []VNIC      `json:"pNICs"`
+	PNICs      []PNIC      `json:"pNICs"`
+	VNICs      []VNIC      `json:"vNICs"`
```
jortel, line 249: Unrelated: Noticed tags were backwards here and fixed.

mansam (approved):

mturley:
@fdupont-redhat @vconzola did we want to also display the subnet mask in the UI on each VMware host network? I think it may have been in an earlier mockup.

fdupont-redhat:
@mturley Without the network mask, showing the network IP address doesn't make sense.
Say that the host IP address is `10.1.0.24`. Without the netmask you don't know if belongs to `10.1.0.0/16` or `10.0.0.0/8`.
From a UI point of view, the slash notation is shorter: `255.255.0.0` becomes `/16`. And it's a common notation.
Most IP address manipulation libraries can switch from one notation to the other.

mturley:
@fdupont-redhat ok thanks, I'll open an issue to follow up on that.

PR #172, opened 2021-02-23
Add Warm Migration support

mansam:
This is the initial implementation of Warm Migration support. This has all the problems with throttling that have been previously discussed, but is a necessary step towards a proper solution. This also doesn't permit changing the cutover datetime while a migration is running. If we want that to be possible, it will need to be done in a follow-up.


pkg/controller/plan/migration.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/konveyor/forklift-controller/pkg/controller/plan/builder"
 	plancontext "github.com/konveyor/forklift-controller/pkg/controller/plan/context"
 	"github.com/konveyor/forklift-controller/pkg/controller/provider/web"
+	"kubevirt.io/vm-import-operator/pkg/apis/v2v/v1beta1"
```
jortel, line 14: Let's import this as `vmio` to follow existing design pattern.

jortel (changes requested):
Overall, LGTM.

jortel (approved):
Thanks

PR #174, opened 2021-02-25
Fix storage and network map validation segfault

mansam:
If a Map's source provider was invalid for a reason other than not being ready, it would attempt to validate the source network or storage and fail because the referenced provider would be nil. This fixes the condition checks to include `SourceProviderNotValid`, which short circuits the validation and prevents the segfault.


jortel (approved):
LGTM

PR #175, opened 2021-02-25
Mark active step complete when VM is canceled

mansam:
This marks the current pipeline step complete when the VM is canceled. Required for https://github.com/konveyor/forklift-ui/pull/400


jortel (approved):
LGTM

PR #177, opened 2021-02-25
Fix mapping CRs reference watch on provider.

jortel:
Having the `ref:` tag on the provider.Pair shadows the fields on provider.Pair.
The ref _mapper_ will not drill into a _struct_ field with the tag.


mansam (approved):
Good catch, LGTM

PR #178, opened 2021-03-01
Add support for migration hooks.

jortel:
Add migration hooks.
Supported for the _PreHook_ and _PostHook_ pipeline steps.
Hook specified per VM/step:
```
- id: vm-2957
  hooks:
    - hook:
        namespace: openshift-migration
        name: test
      step: PreHook
```
Adds hook controller for validation.


pkg/controller/provider/web/ocp/client.go
```diff
@@ -113,6 +113,9 @@ func (r *Finder) With(client base.Client) base.Finder {
 //   NotFoundErr
 //   RefNotUniqueErr
 func (r *Finder) ByRef(resource interface{}, ref base.Ref) (err error) {
+	err = liberr.Wrap(&NotFoundError{
+		Ref: ref,
+	})
```
jortel, line 118: These OCP methods need to return an _error_.  I ran into this by mistake.

mansam (approved):
LGTM

PR #179, opened 2021-03-03
VMIO CR generated names.

jortel:
Convert `KubeVirt` to use generated names.  This includes finding resources by label.
This also includes consolidating the _ensureXX()_ to build the VM Import CR and associated secret.


mansam (approved):
LGTM

PR #181, opened 2021-03-05
Add VM snapshot property.

jortel:
Add VM snapshot property.
```
"snapshot": {
    "kind": "VirtualMachineSnapshot",
    "id": "snapshot-3134"
},
```

Blank when the VM has not current snapshot.
```
"snapshot": {
    "kind": "",
    "id": ""
},
```


pkg/controller/provider/container/vsphere/model.go
```diff
@@ -44,29 +44,38 @@ func (b *Base) Apply(m *model.Base, u types.ObjectUpdate) {
 //
 // Build ref.
 func (b *Base) Ref(in types.AnyType) (ref model.Ref) {
```
jortel, line 46: Refactored to handle `ManagedObjectReferences` pointers as well.

mansam (approved):
LGTM

PR #182, opened 2021-03-08
Provide workload instead of VM in the hook configMap.

jortel:
Provide workload instead of VM in the hook configMap.

Needed to add the `Workload` to the client objects.  The _workload_ is considered a provider non-specific (common) resource.

### DIFF
--- a/pkg/controller/provider/web/base/client.go
+++ b/pkg/controller/provider/web/base/client.go
@@ -79,6 +79,13 @@ type Finder interface {
 	//   NotFoundErr
 	//   RefNotUniqueErr
 	VM(ref *Ref) (interface{}, error)
+	// Find a workload by ref.
+	// Returns the matching resource and:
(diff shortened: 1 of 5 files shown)


fdupont-redhat:
What is the difference between a workload and a VM? Why is the workload notion introduced?

pkg/controller/provider/web/base/client.go
```diff
@@ -135,12 +142,20 @@ type Client interface {
...
+	//   NotFoundErr
+	//   RefNotUniqueErr
+	Workload(ref *Ref) (interface{}, error)
 	// Find a Network by ref.
 	// Returns the matching resource and:
 	//   ProviderNotSupportedErr
 	//   ProviderNotReadyErr
 	//   NotFoundErr
 	//   RefNotUniqueErr
+
```
mansam, line 158: stray whitespace

PR #188, opened 2021-03-16
Use StringData to update secret in ensureSecret

mansam:
Because `builder.Secret` creates a Secret with `StringData` instead of `Data`, `ensureSecret` needs to use the new secret's `StringData` to update the existing secret.


jortel (approved):
LGTM

PR #189, opened 2021-03-16
Don't set a default cutover

mansam:
The UI doesn't schedule a cutover right away, so we need to be able to start a plan with no cutover date.


jortel (approved):
LGTM

PR #190, opened 2021-03-18
Introduce vSphere migration throttling

mansam:
This introduces a stateless scheduling mechanism for vSphere, which restricts the number of VM migrations that can be started across all plans using the same source. 

* Each host can have a number of disks in flight equal to `MAX_VM_INFLIGHT`. VMs can be scheduled if there are enough slots free on the host to accommodate the number of disks attached to the VM. 
* Each Plan can start at most a single VM on each reconcile, in order to allow other Plans an opportunity to schedule a VM.
* `migration.Run` has been refactored to first find all the Plan's started VM migrations and step through their pipelines. Then it acquires the next VM that can be scheduled if any, and steps through its pipeline.

### DIFF
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -105,11 +105,11 @@ func (r Migration) Run() (reQ time.Duration, err error) {
 		}
 	}
 
-	vm, err := r.scheduler.Next()
+	vm, hasNext, err := r.scheduler.Next()
 	if err != nil {
 		return
 	}
-	if vm != nil {
+	if hasNext {
 		err = r.step(vm)
 		if err != nil {
 			return
@@ -124,6 +124,9 @@ func (r Migration) Run() (reQ time.Duration, err error) {
 	return
 }
 
+//
+// Steps a VM through the migration itinerary
+// and updates its status.
 func (r Migration) step(vm *plan.VMStatus) (err error) {
 	// check whether the VM has been canceled by the user
 	if r.Context.Migration.Spec.Canceled(vm.Ref) {
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -276,7 +276,6 @@ func (r *Migration) init() (err error) {
 	}
 	r.scheduler, err = scheduler.New(r.Context)
 	if err != nil {
-		err = liberr.Wrap(err)
 		return
 	}
 
--- a/pkg/controller/plan/migration.go
+++ b/pkg/controller/plan/migration.go
@@ -9,6 +9,7 @@ import (
 	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1alpha1/plan"
 	"github.com/konveyor/forklift-controller/pkg/controller/plan/builder"
 	plancontext "github.com/konveyor/forklift-controller/pkg/controller/plan/context"
+	"github.com/konveyor/forklift-controller/pkg/controller/plan/scheduler"
 	"github.com/konveyor/forklift-controller/pkg/controller/provider/web"
 	vmio "kubevirt.io/vm-import-operator/pkg/apis/v2v/v1beta1"
 	"time"
@@ -70,6 +71,8 @@ type Migration struct {
 	kubevirt KubeVirt
 	// VM import CRs.
 	importMap ImportMap
+	// VM scheduler
+	scheduler scheduler.Scheduler
 }
 
 //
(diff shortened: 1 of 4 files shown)


pkg/controller/plan/migration.go
```diff
@@ -95,119 +98,113 @@ func (r Migration) Run() (reQ time.Duration, err error) {
...
-			if err != nil {
+	completed, err := r.end()
+	if completed {
+		reQ = NoReQ
+	}
+
+	return
+}
+
+func (r Migration) step(vm *plan.VMStatus) (err error) {
```
jortel, line 127: Please add method _doc_ comment.

pkg/controller/plan/scheduler/vsphere/scheduler.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+//
+// Scheduler for migrations from ESX hosts.
+type Scheduler struct {
+	*plancontext.Context
+	MaxInFlight int
+	inflight    map[string]int
+	pending     map[string][]*pendingVM
+}
```
jortel, line 19: Please add fields comments for MaxInFlight, inFlight and pending.

pkg/controller/plan/scheduler/vsphere/scheduler.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// Since we modify the plan VMStatuses in memory,
+	// we need to use the plan from the context rather
+	// than from the list of plans that are retrieved below.
+	for _, vmStatus := range r.Plan.Status.Migration.VMs {
+		obj, err := r.Source.Inventory.VM(&vmStatus.Ref)
+		if err != nil {
+			return err
+		}
+		vm := obj.(*vsphere.VM)
```
jortel, line 58: Since the _ref_ has been resolved during validation, I would recommend using the inventory.Get(&vsphere.VM{}, vmStatus.Ref.ID) would be more consistent with other places in the code and the preferred usage of the client.

pkg/controller/plan/scheduler/vsphere/scheduler.go
```diff
@@ -0,0 +1,117 @@
...
+		if p.Name == r.Plan.Name && p.Namespace == r.Plan.Namespace {
+			continue
+		}
+		// ignore plans that aren't using the same source provider
+		if p.Spec.Provider.Source != r.Plan.Spec.Provider.Source {
+			continue
+		}
+
+		for _, vmStatus := range p.Status.Migration.VMs {
+			obj, err := r.Source.Inventory.VM(&vmStatus.Ref)
```
jortel, line 86: Same comment re: `Find()` here.

pkg/controller/plan/scheduler/vsphere/scheduler.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+type pendingVM struct {
+	status *plan.VMStatus
+	cost   int
+}
+
+//
+// Return the next VM to migrate.
+func (r *Scheduler) Next() (vm *plan.VMStatus, err error) {
```
jortel, line 28: Although there are examples of this in the MT code, I would recommend against using `nil` as = `has next`.
Recommend and explicit: 
```
func (r *Scheduler) Next() (vm *plan.VMStatus, hasNext bool, err error) 
```

Or return a list of VMs that can be started.  The list could be empty.
```
```func (r *Scheduler) Next() (vm []*plan.VMStatus, err error) {

File: pkg/controller/plan/scheduler/vsphere/scheduler.go [hunk 6]
```diff
@@ -0,0 +1,117 @@
+package vsphere
+
+import (
+	"context"
+	liberr "github.com/konveyor/controller/pkg/error"
+	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1alpha1"
+	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1alpha1/plan"
+	plancontext "github.com/konveyor/forklift-controller/pkg/controller/plan/context"
+	"github.com/konveyor/forklift-controller/pkg/controller/provider/web/vsphere"
```
jortel, line 9: One other consideration.  The pattern I have been following is to import the web (or model) resources within provider specific packages as just: _web_ or _model_ for clarity since it's already determined to be provider (type) specific.

pkg/controller/plan/migration.go
```diff
@@ -262,6 +274,11 @@ func (r *Migration) init() (err error) {
 		Context: r.Context,
 		Builder: r.builder,
 	}
+	r.scheduler, err = scheduler.New(r.Context)
+	if err != nil {
+		err = liberr.Wrap(err)
```
jortel, line 279: The liberr.New() used by scheduler.New() will return an already wrapped error.

pkg/controller/plan/scheduler/doc.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	liberr "github.com/konveyor/controller/pkg/error"
+	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1alpha1"
+	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1alpha1/plan"
+	plancontext "github.com/konveyor/forklift-controller/pkg/controller/plan/context"
+	"github.com/konveyor/forklift-controller/pkg/controller/plan/scheduler/vsphere"
+	"github.com/konveyor/forklift-controller/pkg/settings"
+)
+
+type Scheduler interface {
```
jortel, line 12: Please document the interface (and its methods)

jortel (approved):
LGTM

PR #193, opened 2021-03-24
Only PATCH the vmimport cutover if it's been changed

mansam:
The VMI CRs are currently getting updated on every reconcile, which is the likely cause of https://bugzilla.redhat.com/show_bug.cgi?id=1940694. This PR replaces that update with a PATCH, and only if the cutover date has changed. This should clear up the resource conflicts.


jortel (approved):
LGTM

PR #198, opened 2021-03-26
Trim suffix from disk backing file names to fix task name comparisons

mansam:
`buildPipeline` uses the disk backing file name from
the inventory to create the VM pipeline tasks, but
`updatePipeline` uses the disk backing name from the
DataVolumes to find those tasks.

When VMIO creates a DataVolume for a VM that is being warm
migrated, it creates a snapshot first. This will cause
the disk backing file name to differ from what the inventory
saw, and means `updatePipeline` won't be able to find
the task that `buildPipeline` created.

This patch causes the backing file suffixes to be trimmed
before comparing them, enabling pipeline updates to
work correctly.


jortel (approved):
LGTM
Thanks for incorporating into the builder.

PR #201, opened 2021-03-29
Add inventory watch to remaining controllers.

jortel:
Add inventory watch to remaining controllers:
- plan
- map/storage
- host

PR #203, opened 2021-03-31
Provider API: add authorization.

jortel:
Token must have verb="*" on the `Provider` CR to access associated endpoints.
Perform TokenReview/SAR.  The API is hit many times per second so the result is cached for 10 seconds to prevent spamming the API server and for better performance.

Requires: https://github.com/konveyor/controller/pull/57

Also fixes some issues related to watching other forklift resources and a SEGV in the plan controller.  These issues introduced in: 
- https://github.com/konveyor/forklift-controller/pull/201
-  https://github.com/konveyor/forklift-controller/pull/194

### DIFF
--- a/pkg/controller/plan/controller.go
+++ b/pkg/controller/plan/controller.go
@@ -106,7 +106,10 @@ func Add(mgr manager.Manager) error {
 			Type: &api.Provider{},
 		},
 		libref.Handler(),
-		&ProviderPredicate{})
+		&ProviderPredicate{
+			client:  mgr.GetClient(),
+			channel: channel,
+		})
 	if err != nil {
 		log.Trace(err)
 		return err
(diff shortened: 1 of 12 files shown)


pkg/controller/plan/controller.go
```diff
@@ -106,7 +106,10 @@ func Add(mgr manager.Manager) error {
 			Type: &api.Provider{},
 		},
 		libref.Handler(),
-		&ProviderPredicate{})
+		&ProviderPredicate{
+			client:  mgr.GetClient(),
+			channel: channel,
+		})
```
jortel, line 112: Fixes SEGV.

mansam (approved):
LGTM

jortel:
Rebased and added dep: controller v0.3.2 and re-tested.

PR #210, opened 2021-04-12
Fix inventory watch handler for Deleted resources.

jortel:
Fix inventory watch handler for `Deleted` resources.  The logic that resolves the Refs in each CR is flawed because in the case of a `Deleted` resource, the lookup will never be successful which results in the affected CR never being matched.  This logic replaced with _smart_ comparison of the Ref to the deleted/changed resource.  This does not affect OCP resources.

Also, noticed that the predicate should be handling renamed vSphere resources.


mansam (approved):
LGTM

PR #211, opened 2021-04-12
Adjust VM validation search settings and paging.

jortel:
Shorten the validation search interval.  10 minutes seems much more reasonable then 30 seconds.  The watch events should keep up and the search is only needed for in case the watch events queue overflows.

Also, update the search file list VM in pages to limit memory footprint.

### DIFF
--- a/pkg/controller/provider/container/vsphere/watch.go
+++ b/pkg/controller/provider/container/vsphere/watch.go
@@ -67,8 +67,8 @@ type VMEventHandler struct {
 // Search interval.
 func (r *VMEventHandler) searchInterval() time.Duration {
 	seconds := Settings.PolicyAgent.SearchInterval
-	if seconds < 10 {
-		seconds = 10
+	if seconds < 60 {
+		seconds = 60
 	}
 
 	return time.Second * time.Duration(seconds)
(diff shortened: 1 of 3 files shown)


pkg/controller/provider/container/vsphere/watch.go
```diff
@@ -181,25 +181,33 @@ func (r *VMEventHandler) search() {
...
+	for {
+		list := []model.VM{}
+		err = r.DB.List(
+			&list,
+			libmodel.ListOptions{
+				Predicate: libmodel.Or(
+					libmodel.Neq("Revision", libmodel.Field{Name: "RevisionValidated"}),
+					libmodel.Neq("PolicyVersion", version)),
+				Page: &libmodel.Page{
+					Limit: 250,
```
mansam, line 193: I think you might be missing an offset here.

pkg/controller/validation/policy/client.go
```diff
@@ -343,7 +343,7 @@ func (r *Scheduler) Submit(task *Task) (err error) {
 	select {
 	case r.input <- task:
 		// queued.
-	case <-time.After(10 * time.Second):
+	case <-time.After(10 * time.Minute):
```
mansam, line 346: I think the method comment may need to be updated here.

mansam (changes requested):

pkg/controller/provider/container/vsphere/watch.go
jortel: The offset (default to 0) is intentional.  Each pass will validate VM which will be excluded from the next query because validated VMs will not match.

mansam: Ah, understood. Thanks for explaining!

mansam (approved):

PR #212, opened 2021-04-13
Better logging

jortel:
Logging improved.
Few systemic changes for all controllers and added _some_ logging statements.
- The logging.Logger.Reset() removed so added this logic to reconciler.
- Replaced package level logger with a logger on 1st class objects.  Eg: Reconcile.log.  The Reconciler.Reconcile() method updated to be non-pointer receiver.  This ensures the logger is not shared between _workers_.
- The logging.Logger no longer filters out k8s `Conflict` errors.  Each reconciler needs to do it instead.

Updated the _reference_ watch event handler to be configured with the controller's CR as the _owner_.  This is required to comply with the new signature and to support proper event filtering.

Requires: https://github.com/konveyor/controller/pull/63

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -17,8 +17,10 @@ limitations under the License.
 package main
 
 import (
+	"github.com/konveyor/controller/pkg/logging"
 	"github.com/konveyor/forklift-controller/pkg/settings"
 	"os"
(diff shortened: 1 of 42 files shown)


mansam (approved):
LGTM

PR #213, opened 2021-04-13
VM should reflect Pending condition from VM import

mansam:
* Make VM status reflect the VMI Pending condition
* If a DV hasn't bound yet, reflect the reason why in the DiskTransfer step's phase.


jortel (approved):
LGTM

PR #218, opened 2021-04-16
Skip VMs that are not connected

### LINKED ISSUE #138: Validation - Block VMware VM in inaccessible state
When a VM is in inaccessible state it will not report an operating system and the migration will fail, but the message is wrong.
It would be clearer to detect that the VM is inaccessible and block the migration at the validation stage.

The attribute to identify the state is `.runtime.connectionState` and the value can be `connected`, `disconnected`, `orphaned`, `invalid` or `inaccessible`. The import must fail when the value is not `connected`.

This information should also be in the inventory to allow the UI to display a warning.

fdupont-redhat:
If a VMware VM connection state is not "Connected", the migration will fail. This pull request adds the `ConnectionState` attribute for the VM in the VMware inventory and errors out when building the VirtualMachineImport CR if the value is not `VirtualMachineConnectionStateConnected`.

Fixes #138.

### DIFF
--- a/pkg/controller/plan/builder/vsphere/builder.go
+++ b/pkg/controller/plan/builder/vsphere/builder.go
@@ -100,6 +100,13 @@ func (r *Builder) Import(vmRef ref.Ref, object *vmio.VirtualMachineImportSpec) (
 				pErr.Error()))
 		return
 	}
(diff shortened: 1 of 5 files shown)


jortel (approved):
LGTM
Very nice PR, @fdupont-redhat !

fdupont-redhat:
Tested with `make run`. The `connectionState` attribute is correctly set.

PR #223, opened 2021-04-20
[BZ 1951395] Warm Migration - Skip VMs with CBT disabled

fdupont-redhat:
When a plan is set to be warm, the VM needs to have Changed Block Tracking (CBT) enabled, in order to identify the delta between two snapshots. Without this feature, the warm migration will not work.

In VMIO, the check for CBT will raise a validation error and requeue, which leaves the migration stuck from a Forklift perspective. So, we decided to add a check in the plan migration builder to fail VMs with CBT disabled directly in the controller. This will avoid creating a VirtualMachineImport CR that will be stuck.

This pull request implements this new behavior.


jortel (approved):
LGTM

Issue #103, opened 2020-12-07
Validate all of the VMs are covered by the mapped resources.


jortel:
Validate all of the VMs are covered by the mapped resources.  The validation needs to iterate the VMs and lookup the network and datastore in the mapping.  Reflect deficiencies in a condition.

PR #226, opened 2021-04-20
Builder -> Adapter refactor; VM Map validation

mansam:
* Created a Refs collection type and added it to the MapStatus.
* Resolved Network and Storage refs are stored in MapStatus during validation.
* Added `Adapter` type which is responsible for creating `Builders` and `Validators`
* Moved `Builder` under the `Adapter` package
* Added `Validator` type which is responsible for provider-specific validation behavior
* Added mapping validation to Plan controller in validateVM


pkg/controller/plan/adapter/vsphere/validator.go
```diff
@@ -0,0 +1,70 @@
...
+func (r *Validator) Load() (err error) {
+	r.inventory, err = web.NewClient(r.plan.Referenced.Provider.Source)
+	return
+}
+
+//
+// Validate that a VM's networks have been mapped.
+func (r *Validator) NetworksMapped(vmRef ref.Ref) (ok bool, err error) {
+	vm := &model.VM{}
+	pErr := r.inventory.Find(vm, vmRef)
```
jortel, line 30: This _can_ assign `err` directly and it would be simpler.

Also, rather than creating a new error, you can wrap `err` with more context.
Eg:
```
err = liberr.Wrap(
    err,
    "VM not found."
   "vm",
  vmRef.String())
```
The example uses _structural_ logging instead of the sprintf.

jortel (approved):
Looks great!

fdupont-redhat:
Fixes #103

PR #229, opened 2021-04-23
Ensure vsphere & esx host connections are completely closed.

jortel:
Ensure vsphere & esx host connections are completely closed.


mansam (approved):
LGTM!

PR #233, opened 2021-04-26
Fix HTTP client connection leak.

jortel:
Fix HTTP client connection leak.
The default number of idle connections (and idle timeout) for `http.Transport` is unlimited. This patch limits both to ensure that connections associated with Transports that go out of scope will get closed.


mansam (approved):
LGTM

PR #236, opened 2021-04-27
BZ 1952312 - Add VM name to builder error messages

fdupont-redhat:
When the VirtualMachineImport builder fails, the VM MoRef is not user friendly in the error. This pull request adds the VM name.

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -95,8 +95,7 @@ func (r *Builder) Import(vmRef ref.Ref, object *vmio.VirtualMachineImportSpec) (
 	if pErr != nil {
 		err = liberr.New(
 			fmt.Sprintf(
-				"VM %s (%s) lookup failed: %s",
-				vm.Name,
+				"VM %s lookup failed: %s",
 				vmRef.String(),
 				pErr.Error()))
 		return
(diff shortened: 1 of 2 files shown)


pkg/controller/plan/adapter/vsphere/builder.go
```diff
@@ -95,22 +95,25 @@ func (r *Builder) Import(vmRef ref.Ref, object *vmio.VirtualMachineImportSpec) (
 	if pErr != nil {
 		err = liberr.New(
 			fmt.Sprintf(
-				"VM %s lookup failed: %s",
+				"VM %s (%s) lookup failed: %s",
+				vm.Name,
 				vmRef.String(),
```
jortel, line 100: We use the Ref.String() everywhere for both error and logging.  Let's change it to provide better information.
Perhaps:
```
//
// String representation.
// Format: (<type>) id: <id> name: '<name>'
func (r *Ref) String() (s string) {
	if r.Type != "" {
		s = fmt.Sprintf(
			"(%s)",
			r.Type)
	}
	s = fmt.Sprintf(
		"%s id:%s name:'%s' ",
		s,
		r.ID,
		r.Name)
	
	return
}
```

jortel (changes requested):
Recommend _centralized_ solution.

jortel (approved):
LGTM

PR #237, opened 2021-04-27
API v1alpha1 to: v1beta1 for GA.

jortel:
API v1alpha1 to: v1beta1 for GA.
Discussed with @shawn-hurley that since we don't plan to support both v1alpha1 and v1beta1, we should just replace the package and regenerate the CRD instead of adding v1beta1.


mansam (approved):
LGTM

PR #239, opened 2021-04-27
Fix SEGV in vSphere validator when Maps are missing

mansam:


pkg/controller/plan/adapter/vsphere/validator.go
```diff
@@ -25,6 +25,10 @@ func (r *Validator) Load() (err error) {
 //
 // Validate that a VM's networks have been mapped.
 func (r *Validator) NetworksMapped(vmRef ref.Ref) (ok bool, err error) {
+	if r.plan.Referenced.Map.Network == nil {
+		ok = true
```
jortel, line 29: Shouldn't this be `false`.  No map means every VM isn't mapped.

jortel (approved):
LGTM

PR #240, opened 2021-04-27
Make TransferNetwork field an ObjectReference

mansam:
When the `v1.multus-cni.io/default-network` annotation is set with the name of a network attachment definition, it looks for that NAD in `kube-system`, not in the pod's namespace as I previously thought. It's also valid in some circumstances to choose a NAD in a namespace outside of the pod's namespace. As such, the `transferNetwork` field on the Plan needs to be an object reference to properly identify the desired network.

This requires a corresponding change to the UI which I've already discussed with @mturley. 

https://github.com/k8snetworkplumbingwg/multus-cni/blob/a52680b3bfa026087879f89fae331b0dfc8dffe0/docs/configuration.md

### DIFF
--- a/pkg/apis/forklift/v1beta1/plan.go
+++ b/pkg/apis/forklift/v1beta1/plan.go
@@ -21,6 +21,7 @@ import (
 	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/plan"
 	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/provider"
 	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/ref"
+	core "k8s.io/api/core/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/apis/forklift/v1beta1/plan.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/plan"
 	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/provider"
 	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/ref"
+	v1 "k8s.io/api/core/v1"
```
jortel, line 24: please alias as `core` for consistency.

pkg/controller/plan/kubevirt.go
```diff
@@ -327,8 +328,9 @@ func (r *KubeVirt) vmImport(
 		return
 	}
 	annotations := make(map[string]string)
-	if r.Plan.Spec.TransferNetwork != "" {
-		annotations[annDefaultNetwork] = r.Plan.Spec.TransferNetwork
+	if r.Plan.Spec.TransferNetwork != nil {
+		annotations[annDefaultNetwork] = fmt.Sprintf(
+			"%s/%s", r.Plan.Spec.TransferNetwork.Namespace, r.Plan.Spec.TransferNetwork.Name)
```
jortel, line 333: In may other places, I have use path.Join() instead because it's more concise.

jortel (approved):
LGTM

mturley:
Corresponding UI change: https://github.com/konveyor/forklift-ui/pull/559

PR #250, opened 2021-05-05
Scheduler improvements

mansam:
The scheduler should ignore Plans that aren't running, and it shouldn't bother looking up a VM in the inventory unless that VM is being migrated. For safety's sake VMs that can't be found in the inventory are also ignored.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1956467


jortel (approved):
LGTM

fdupont-redhat (approved):

PR #252, opened 2021-05-06
BZ 1957105 - Skip VMs that are actually templates

fdupont-redhat:
In VMware, a template is a VM with a boolean property that says it is a template. The property is `config.template`.
Templates don't support warm migration, so we have decided to hide all the templates from the VM selection list in the UI.

This pull request does two things:
1. add a new field in the inventory: `IsTemplate` in the database, `isTemplate` is the JSON API response.
2. error out the templates in the VMImport builder, so that their migration is skipped.

Signed-off-by: Fabien Dupont <fdupont@redhat.com>


jortel (approved):
LGTM

fdupont-redhat:
I've tested it in my lab and the inventory shows the `isTemplate` field with the correct value.

PR #256, opened 2021-05-10
Remove the cutover field from the Plan CRD

fdupont-redhat:
In the migration flow, the cutover date/time is set on the Migration CR by the user interface. We expect user to do the same through the API to keep a consistent flow. This means that the `cutover` field of the `Plan` is not needed anymore. Since we will soon release Forklift 2.0.0 with the API in v1beta1 version, it would be cleaner to do it soon, even if we add it back in the future.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -362,8 +362,6 @@ func (r *KubeVirt) vmImport(
 		object.Spec.Warm = true
 		if r.Migration.Spec.Cutover != nil {
 			object.Spec.FinalizeDate = r.Migration.Spec.Cutover
-		} else {
-			object.Spec.FinalizeDate = r.Plan.Spec.Cutover
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -362,8 +362,6 @@ func (r *KubeVirt) vmImport(
 		object.Spec.Warm = true
 		if r.Migration.Spec.Cutover != nil {
```
mansam, line 363: Please remove this nil check; now that the fallback to the Plan value is gone we need to set `object.Spec.FinalizeData = r.Migration.Spec.Cutover` in all cases so that the user can clear the field if they make a mistake.

```
	if r.Plan.Spec.Warm {
		object.Spec.Warm = true
		object.Spec.FinalizeDate = r.Migration.Spec.Cutover
		
	}
```

mansam (changes requested):

jortel:
LGTM

PR #261, opened 2021-05-12
Fix pointer issue w/ Plan VM ref resolution

mansam:
The refs in the Plan spec's `vms` field weren't being resolved properly due to the ref being a loop copy of the one on the plan, rather than the ref itself. This was causing VMs to not match the list of VMs in the Plan's copy of the migration status, which resulted in completed VMs being restarted.


jortel (approved):
Good catch.

PR #264, opened 2021-05-16
General hardening and memory leak fixes.

jortel:
General hardening and memory leak fixes.
- [ ] https://github.com/konveyor/controller/pull/74 needs to be merged and released.
- [ ] go.mod updated with _controller_ release.

### DIFF
--- a/pkg/controller/host/predicate.go
+++ b/pkg/controller/host/predicate.go
@@ -90,8 +90,8 @@ func (r *ProviderPredicate) Update(e event.UpdateEvent) bool {
 func (r *ProviderPredicate) Delete(e event.DeleteEvent) bool {
 	p, cast := e.Object.(*api.Provider)
 	if cast {
-		reconciled := p.Status.ObservedGeneration == p.Generation
-		return reconciled
+		r.WatchManager.Deleted(p)
+		return true
 	}
 
 	return false
--- a/pkg/controller/provider/container/vsphere/watch.go
+++ b/pkg/controller/provider/container/vsphere/watch.go
@@ -254,7 +254,7 @@ func (r *VMEventHandler) validated(task *policy.Task) {
 		_ = tx.End()
 	}()
 	latest := &model.VM{Base: model.Base{ID: task.Ref.ID}}
-	err = r.DB.Get(latest)
+	err = tx.Get(latest)
 	if err != nil {
 		r.log.Error(err, "get vm failed.")
 		return
--- a/pkg/controller/provider/controller.go
+++ b/pkg/controller/provider/controller.go
@@ -70,7 +70,7 @@ func Add(mgr manager.Manager) error {
 	web.TLS.Enabled = Settings.Inventory.TLS.Enabled
 	web.TLS.Certificate = Settings.Inventory.TLS.Certificate
 	web.TLS.Key = Settings.Inventory.TLS.Key
-	web.AllowedOrigins = Settings.CORS.AllowedOrigins
+	web.AllowedOrigins = Settings.Inventory.CORS.AllowedOrigins
 	reconciler := &Reconciler{
 		Reconciler: base.Reconciler{
 			EventRecorder: mgr.GetEventRecorderFor(Name),
--- a/pkg/controller/validation/policy/client.go
+++ b/pkg/controller/validation/policy/client.go
@@ -15,6 +15,7 @@ import (
 	"net"
 	"net/http"
 	liburl "net/url"
+	"runtime"
 	"time"
 )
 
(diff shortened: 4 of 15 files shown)


mansam (approved):
LGTM

pkg/controller/host/predicate.go
```diff
@@ -90,8 +90,8 @@ func (r *ProviderPredicate) Update(e event.UpdateEvent) bool {
 func (r *ProviderPredicate) Delete(e event.DeleteEvent) bool {
 	p, cast := e.Object.(*api.Provider)
 	if cast {
-		reconciled := p.Status.ObservedGeneration == p.Generation
-		return reconciled
```
jortel, line 94: Noticed this while troubleshooting other issues.  We need to process `Deleted` events regardless of whether the resource is reconciled (or not).

This applies to similar changes elsewhere.

pkg/controller/host/predicate.go
```diff
@@ -90,8 +90,8 @@ func (r *ProviderPredicate) Update(e event.UpdateEvent) bool {
...
+		r.WatchManager.Deleted(p)
```
jortel, line 93: The WatchManager was leaking `Watch` when providers are deleted.

See this change in many other predicates.

pkg/controller/provider/container/vsphere/watch.go
```diff
@@ -254,7 +254,7 @@ func (r *VMEventHandler) validated(task *policy.Task) {
 		_ = tx.End()
 	}()
 	latest := &model.VM{Base: model.Base{ID: task.Ref.ID}}
-	err = r.DB.Get(latest)
+	err = tx.Get(latest)
```
jortel, line 257: Only the transaction should be used while _inside_ a transaction.

pkg/controller/provider/controller.go
```diff
@@ -70,7 +70,7 @@ func Add(mgr manager.Manager) error {
 	web.TLS.Enabled = Settings.Inventory.TLS.Enabled
 	web.TLS.Certificate = Settings.Inventory.TLS.Certificate
 	web.TLS.Key = Settings.Inventory.TLS.Key
-	web.AllowedOrigins = Settings.CORS.AllowedOrigins
+	web.AllowedOrigins = Settings.Inventory.CORS.AllowedOrigins
```
jortel, line 73: Clarity.

pkg/controller/validation/policy/client.go
```diff
@@ -307,10 +315,10 @@ type Worker struct {
 // Process input queue. Validation delegated to the
 // policy agent.
 func (r *Worker) run() {
-	defer func() {
-		_ = recover()
-	}()
 	go func() {
+		defer func() {
+			_ = recover()
+		}()
```
jortel, line 321: The `recover()` needs to be deferred in the function to prevent the panic if the channel is closed.

pkg/controller/watch/handler/watch.go
```diff
@@ -63,3 +63,18 @@ func (m *WatchManager) Ensure(
...
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+	if watchMap, found := m.providerMap[provider.UID]; found {
+		for _, w := range watchMap {
+			w.End()
+		}
+	}
+
+	delete(m.providerMap, provider.UID)
+}
```
jortel, line 80: Called when a provider is deleted to _clean up_ associated cached watches.  Fixes memory leak.

pkg/settings/policy.go
```diff
@@ -43,11 +43,11 @@ func (r *PolicyAgent) Load() (err error) {
 	} else {
 		r.CA = ServiceCAFile
 	}
-	r.Limit.Worker, err = getEnvLimit(PolicyAgentWorkerLimit, 25)
+	r.Limit.Worker, err = getEnvLimit(PolicyAgentWorkerLimit, 10)
```
jortel, line 46: In the current design, this throttles concurrent validation request (validation service) for each provider.  As a result, if there are 10 providers, MTV will be sending 250 concurrent validation requests.  Reducing this because I have seen "broken pipe" errors in the log when the provider inventory API starts logging 2-8 seconds response times.
This is a short term fix.  The long term fix is to:
- Perform policy validation using an internal OPA.
- Rate limit the validation request across all providers.
- Rate limit incoming API requests.

pkg/settings/policy.go
```diff
@@ -43,11 +43,11 @@ func (r *PolicyAgent) Load() (err error) {
...
 	if err != nil {
 		return err
 	}
-	r.Limit.Backlog, err = getEnvLimit(PolicyAgentBacklogLimit, 250)
+	r.Limit.Backlog, err = getEnvLimit(PolicyAgentBacklogLimit, 1024)
```
jortel, line 50: The backlog has a small memory footprint so this can be bigger.  The number is somewhat arbitrary anyway.

PR #265, opened 2021-05-18
Improved policy agent integration.

jortel:
Revised policy agent (validation service) integration.

- Provides concurrency (and throttling) limit across all providers using a shared worker pool.
- Batch updating VM with validated result into into transactions.
- Support cancel of scheduled VM validations when provider is removed.  Immediate, cleaner shutdown.
- Tuned settings.

Some minor logging adjustments.

This is critical to supporting very large Providers.


mansam (approved):
LGTM

PR #266, opened 2021-05-18
Fix logic error w/ kept VMs in Migration.begin()

mansam:
Slice creation and assignment for kept VMs was happening inside of the for loop, so the slice of kept VMs would only ever have a single VM in it. Moved the creation and assignment outside of the loop to fix this.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1953145
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1959731


jortel (approved):
Nice find.

PR #268, opened 2021-05-19
Optimized vSphere tree building.

jortel:
Update the algorithm for building the vSphere trees.  Results in a **5x** improvement in performance and (likely) reduced CPU usage.

**Instead of** traversing each node: following child refs and Get() each child by ref.  The number of queries is a multiple of the number of children of each node.  So 4k VMs results in 4k+ queries.

**The optimization** traverse each node: List() children by parent=_node_.  The number of queries is a multiple of the tree depth.

To achieve this:
- The vSphere `Model` objects needed a few _new_ _natural_ FK to parents.
- The tree _navigators_ needed to be rewritten (they are simple).
- The web resources updated to reflect the new FK fields (mainly to make visual testing easier).
- The existing` VM.Host` needed to be changed from type `Ref` to `string` to support DB query.  The ref is stored has json.
- The vSphere `Builder` and `Scheduler` needed to be updated for the VM.Host field change.

### DIFF
--- a/pkg/controller/plan/adapter/vsphere/builder.go
+++ b/pkg/controller/plan/adapter/vsphere/builder.go
@@ -270,7 +270,7 @@ func (r *Builder) hostID(vmRef ref.Ref) (hostID string, err error) {
 		return
 	}
 
-	hostID = vm.Host.ID
+	hostID = vm.Host
 
 	return
 }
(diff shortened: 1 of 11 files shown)


mansam (approved):
LGTM

PR #272, opened 2021-05-25
Workload passed to the policy agent.

jortel:
Workload passed to the policy agent.  This eliminates the PA calling back to the provider API.

Note: Requires change in forklift-validation.


mansam (approved):
LGTM

pkg/controller/validation/policy/client.go
```diff
@@ -178,7 +166,7 @@ func (r *Client) post(path string, in interface{}, out interface{}) (err error)
 		"url",
 		url,
 		"body",
-		out)
+		in)
```
jortel, line 169: Meant to include the posted document in the log.

PR #274, opened 2021-05-30
Better handling of provider inventory not-ready.

jortel:
Better handling of provider inventory not-ready.
To distinguish between 404 because the provider has not yet been added to the container vs. the resource is not in the inventory, the client will:
- **Currently**: The client fetches the provider on every request.
- **This PR**: The web handler (server) sets the `X-Provider` header when the requested _provider_ is found in the container. If the header is not set, the client determines that the 404 is because the provider has not yet been loaded and returns a `ProviderNotReady` error.  Else, the provider is in the container but the resource is not found.

This enhancement will cut the number of GET requests to the API (from the controller) in half.

Also, noticed the `PartialContent` (206) was not being handled.

**Note**: Requires: https://github.com/konveyor/controller/pull/77

### DIFF
--- a/pkg/controller/host/handler/vsphere/handler.go
+++ b/pkg/controller/host/handler/vsphere/handler.go
@@ -49,9 +49,6 @@ func (r *Handler) Watch(watch *handler.WatchManager) (err error) {
 //
 // Resource created.
 func (r *Handler) Created(e libweb.Event) {
-	if !r.HasParity() {
-		return
-	}
 	if host, cast := e.Resource.(*vsphere.Host); cast {
 		r.changed(host)
 	}
@@ -60,9 +57,6 @@ func (r *Handler) Created(e libweb.Event) {
 //
 // Resource created.
 func (r *Handler) Updated(e libweb.Event) {
-	if !r.HasParity() {
-		return
-	}
 	if host, cast := e.Resource.(*vsphere.Host); cast {
 		updated := e.Updated.(*vsphere.Host)
 		if updated.Path != host.Path {
@@ -74,9 +68,6 @@ func (r *Handler) Updated(e libweb.Event) {
 //
 // Resource deleted.
 func (r *Handler) Deleted(e libweb.Event) {
-	if !r.HasParity() {
-		return
-	}
 	if host, cast := e.Resource.(*vsphere.Host); cast {
 		r.changed(host)
 	}
--- a/pkg/controller/provider/container/vsphere/watch.go
+++ b/pkg/controller/provider/container/vsphere/watch.go
@@ -101,6 +101,9 @@ func (r *VMEventHandler) Started(uint64) {
 // This is best-effort.  If the validate() fails, it wil be
 // picked up in the next search().
 func (r *VMEventHandler) Created(event libmodel.Event) {
+	if r.canceled() {
+		return
+	}
 	if vm, cast := event.Model.(*model.VM); cast {
 		if !vm.Validated() {
 			if r.validate(vm) == nil {
@@ -116,6 +119,9 @@ func (r *VMEventHandler) Created(event libmodel.Event) {
 // This is best-effort.  If the validate() fails, it wil be
 // picked up in the next search().
 func (r *VMEventHandler) Updated(event libmodel.Event) {
+	if r.canceled() {
+		return
+	}
 	if vm, cast := event.Updated.(*model.VM); cast {
 		if !vm.Validated() {
 			if r.validate(vm) == nil {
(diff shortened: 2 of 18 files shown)


pkg/controller/host/handler/vsphere/handler.go
```diff
@@ -49,9 +49,6 @@ func (r *Handler) Watch(watch *handler.WatchManager) (err error) {
 //
 // Resource created.
 func (r *Handler) Created(e libweb.Event) {
-	if !r.HasParity() {
-		return
-	}
```
jortel, line 54: HasParity() no longer needs to be checked because the watch has been setup without the initial snapshot (list of model `Created` events).  The parity marker was used to _skip_ the initial `Created` events.

pkg/controller/provider/container/vsphere/watch.go
```diff
@@ -161,6 +160,7 @@ func (r *VMEventHandler) run() {
 	defer r.log.Info("Run stopped.")
 	interval := time.Second * time.Duration(
 		Settings.PolicyAgent.SearchInterval)
+	r.list()
```
jortel, line 163: The watch is setup with snapshot=false.  List and validate all of the VMs.  Then, go into the scheduled follow up searches.

pkg/controller/provider/container/vsphere/watch.go
```diff
@@ -274,11 +274,7 @@ func (r *VMEventHandler) validate(vm *model.VM) (err error) {
 	}
 	err = policy.Agent.Submit(task)
 	if err != nil {
-		if errors.As(err, &policy.BacklogExceededError{}) {
```
jortel, line 277: The `BacklogExceededError` no longer exists.  The channel is no longer limited because the `Task` is super small so the limit just causes trouble.

pkg/controller/provider/web/base/handler.go
```diff
@@ -84,25 +90,26 @@ func (h *Handler) Link(path string, params Params) string {
...
+			UID: types.UID(uid),
 		},
 	}
 	if h.Provider.UID != "" {
 		if h.Reconciler, found = h.Container.Get(h.Provider); !found {
-			return http.StatusNotFound
+			status = http.StatusNotFound
+			return
 		}
+		ctx.Header(ProviderHeader, uid)
```
jortel, line 107: This is part of the "Better handling of provider not ready."

The web endpoint handler sets `X-Provider` so the client can distinguish between the provider not found (404) and the resource within the provider not found (404).

pkg/controller/provider/web/client.go
```diff
@@ -105,25 +103,9 @@ func (r *ProviderClient) Finder() Finder {
 //   ProviderNotReadyErr
 //   NotFoundErr
 func (r *ProviderClient) Get(resource interface{}, id string) (err error) {
-	err = r.find()
```
jortel, line 108: Don't attempt to fetch the provider resource first.

pkg/controller/provider/web/client.go
```diff
@@ -136,25 +118,9 @@ func (r *ProviderClient) Get(resource interface{}, id string) (err error) {
 //   ProviderNotReadyErr
 //   NotFoundErr
 func (r *ProviderClient) List(resource interface{}, param ...Param) (err error) {
-	err = r.find()
```
jortel, line 139: Don't attempt to fetch the provider resource first and examine the X-Provider header instaed.
Smarter and reduces API requests.

pkg/controller/provider/web/client.go
```diff
@@ -167,25 +133,9 @@ func (r *ProviderClient) List(resource interface{}, param ...Param) (err error)
 //   ProviderNotReadyErr
 //   NotFoundErr
 func (r *ProviderClient) Watch(resource interface{}, h EventHandler) (w *Watch, err error) {
-	err = r.find()
```
jortel, line 170: Don't attempt to fetch the provider resource first and examine the X-Provider header instaed.
Smarter and reduces API requests.

pkg/controller/validation/policy/client.go
```diff
@@ -367,8 +358,8 @@ func (r *Pool) Start() {
 	if r.started {
 		return
 	}
-	r.output = make(chan *Task, r.backlog())
-	r.input = make(chan *Task, r.backlog())
+	r.output = make(chan *Task)
+	r.input = make(chan *Task)
```
jortel, line 362: The `Task` object is super small and limiting the channel just causes problems.

pkg/controller/watch/handler/handler.go
```diff
@@ -71,13 +69,6 @@ func (r *Handler) Match(object meta.Object, ref core.ObjectReference) bool {
 		ref.Name == object.GetName()
 }
 
-//
-// Inventory watch has parity.
-func (r *Handler) HasParity() bool {
-	return r.parity
-
-}
```
jortel, line 79: No longer need to keep track of when parity is reached.  That watch is created with snapshot=false.

mansam (approved):
LGTM

PR #276, opened 2021-06-02
Additional oVirt fields.

jortel:
_"Added a ton of properties."_

Add mainly, VM, Host, NIC and Disk fields.
Add DiskProfile.
Add revision management and concerns support.
update `workload`.


mansam (approved):
LGTM

PR #277, opened 2021-06-07
Refactor Handler.Link() for better separation of responsibility and re-use outside the web layer.

jortel:
Refactor Handler.Link() for better separation of responsibility and re-use both inside and outside the web layer.
It's useful to Link and Expand REST resources outside the web layer.  It's complex and potentially dangerous to artificially build and use the _Handler_ because of unknown Handler state needs.

The main driver is to support the VM validation building the workload which needs to be a REST resource.

Essentially, functionality moved from endpoint _Handler_ to REST resources.

Summary:
- Removed Handler.Link().
- Added _Resource_.Link().
- Removed Handler.Build().
- Added _Resource_.Expand() to resources as necessary.  Here the term _expand_ refers to de-referencing foreign keys.

For consistency, we'll want to do the same refactor for vSphere.

### DIFF
--- a/pkg/controller/provider/web/base/handler.go
+++ b/pkg/controller/provider/web/base/handler.go
@@ -34,6 +34,18 @@ const (
 type Params = map[string]string
 
 //
+// Build link.
+func Link(path string, params Params) string {
+	for k, v := range params {
+		if len(v) > 0 {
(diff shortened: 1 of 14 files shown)


mansam (approved):
LGTM

PR #280, opened 2021-06-08
oVirt Builder and Scheduler

mansam:
* Adds ProvisionedSize field to the Disk
* Implements the oVirt builder. The secret requires a `cacert` field containing a b64 encoded root CA.
* Implements the oVirt scheduler. The scheduler will allow up to MaxInFlight VMs per provider to be running at once. We should probably have vSphere and oVirt have their own settings for the maximum number of transfers in flight, since vSphere is disks not VMs, and in any case limitations will be different per provider. This can be done in a follow up.


jortel (approved):
LGTM

PR #282, opened 2021-06-14
Load() the oVirt builder while constructing it

mansam:
The oVirt Builder wasn't having `Load()` called on it before, which meant the provisioner map was unpopulated and the destination storage would not have the correct volume and access modes set.


jortel (approved):
LGTM

PR #284, opened 2021-06-16
Revert "Don't reconcile successful plans"

mansam:
Since we are always permitting retries in 2.1, this can be reverted.

Reverts konveyor/forklift-controller#259


jortel (approved):
LGTM
We should add task _closing_ soon to terminate reconciling plans after users are finished with them.

PR #286, opened 2021-06-18
controller 0.4.3 compat.

jortel:
The `Model` interface related.  String(), Equals() and Labels() now optional.
Using `sql:incremented` instead of Created() and Updated() methods on the models.
Detail level=1 no longer means ALL.
The PK no longer blindly derived when not set.  As a result, the OCP primary key is the UID which is more straight forward anyway.

**Requires**: https://github.com/konveyor/controller/pull/82 expected to be released as 0.4.3.  Will update the PR when it's released.

### DIFF
--- a/pkg/controller/provider/model/ocp/model.go
+++ b/pkg/controller/provider/model/ocp/model.go
@@ -20,6 +20,10 @@ var NotFound = libmodel.NotFound
 
 type InvalidRefError = base.InvalidRefError
 
+const (
+	MaxDetail = base.MaxDetail
+)
+
 //
 // Types
 type Model = base.Model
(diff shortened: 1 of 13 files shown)


pkg/controller/provider/model/ocp/model.go
```diff
@@ -36,10 +40,8 @@ type Resource interface {
 //
 // Base k8s model.
 type Base struct {
-	// PK
-	PK string `sql:"pk"`
 	// Object UID.
-	UID string `sql:"d0"`
+	UID string `sql:"pk"`
```
jortel, line 44: Let's just use the UID as the PK.

mansam (approved):

PR #287, opened 2021-06-21
Improved oVirt (change) detection and event handling.

jortel:
Add additional _const_ for event codes.

Disk.StorageUsed renamed: `ActualSize` per RHV native naming.

A support for DataCenter events though don't know how to test it.  Not sure how to add/update/delete a DC.

Add support for vNICProfile events.  This uses the new container.Collection since these events do not indicate which profile changed.  A watch is added that will determine which VMs are affected and _tag_ them for re-validation.

This PR established this pattern.

### DIFF
--- a/pkg/controller/provider/container/ovirt/model.go
+++ b/pkg/controller/provider/container/ovirt/model.go
@@ -3,6 +3,7 @@ package ovirt
 import (
 	liberr "github.com/konveyor/controller/pkg/error"
 	fb "github.com/konveyor/controller/pkg/filebacked"
+	libcnt "github.com/konveyor/controller/pkg/inventory/container"
 	libmodel "github.com/konveyor/controller/pkg/inventory/model"
 	libweb "github.com/konveyor/controller/pkg/inventory/web"
 	model "github.com/konveyor/forklift-controller/pkg/controller/provider/model/ovirt"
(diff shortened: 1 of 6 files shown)


pkg/controller/provider/container/ovirt/model.go
```diff
@@ -376,6 +458,10 @@ func (r *ClusterAdapter) Apply(client *Client, tx *libmodel.Tx, event *Event) (e
 		m := &model.Cluster{
 			Base: model.Base{ID: object.ID},
 		}
+		err = tx.Get(m)
+		if err != nil {
+			break
+		}
```
jortel, line 464: The Get() was missing everywhere.

pkg/controller/provider/container/ovirt/resource.go
```diff
@@ -601,7 +601,7 @@ type Disk struct {
 		List []Ref `json:"storage_domain"`
 	} `json:"storage_domains"`
 	Status      string `json:"status"`
-	StorageUsed string `json:"actual_size"`
+	ActualSize  string `json:"actual_size"`
```
jortel, line 604: More _native_ naming.

mansam (approved):
LGTM

PR #289, opened 2021-06-22
Host maintenance mode validations

mansam:
* Update the Host in the inventory if its InMaintenanceMode value has changed.
* Add a warning condition to the Host CR if it's in maintenance mode.
* Add a warning condition to the Plan CR if any of its VMs are on hosts which are in maintenance mode.


jortel (approved):
LGTM

PR #291, opened 2021-06-23
Let VMIO manage VM start during warm migration

mansam:
Lets VMIO handle whether or not to start a VM during a warm migration. This permits it to use the VM's powerstate immediately before cutover to determine whether or not to start the destination VM. 

https://bugzilla.redhat.com/show_bug.cgi?id=1957791


jortel (approved):
LGTM

PR #294, opened 2021-06-24
Fix Plan.TargetNamespace() nil pointer dereference

mansam:
Plan.TargetNamespace() was falling back to the `Referenced` Plan's `Namespace` field. The `Referenced` Plan field is not populated for Plans, so this fails.

### DIFF
--- a/pkg/controller/plan/validation.go
+++ b/pkg/controller/plan/validation.go
@@ -22,6 +22,7 @@ import (
 //
 // Types
 const (
+	NamespaceNotValid   = "NamespaceNotValid"
 	TransferNetNotValid = "TransferNetworkNotValid"
 	NetRefNotValid      = "NetworkMapRefNotValid"
 	NetMapNotReady      = "NetworkMapNotReady"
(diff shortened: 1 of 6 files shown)


jortel (approved):
LGTM

jortel:
I think we agreed that part of this fix needs to be to remove the `omitempty` from the targetNamespace to make it required.

pkg/controller/plan/validation.go
```diff
@@ -125,6 +134,27 @@ func (r *Reconciler) validate(plan *api.Plan) error {
...
+// Validate the target namespace.
+func (r *Reconciler) validateTargetNamespace(plan *api.Plan) (err error) {
+	err = r.Get(context.TODO(), types.NamespacedName{Name: plan.Spec.TargetNamespace}, &v1.Namespace{})
+	if k8serr.IsNotFound(err) {
+		err = nil
+		newCnd := libcnd.Condition{
+			Type:     NamespaceNotValid,
+			Status:   True,
+			Category: Critical,
+			Reason:   NotFound,
```
jortel, line 147: I think `NotSet` would be better.

jortel (approved):
LGTM

pkg/controller/plan/validation.go
mansam: Maybe NotSet if it's empty, NotFound if the namespace doesn't exist?

pkg/controller/plan/validation.go
```diff
@@ -125,6 +132,27 @@ func (r *Reconciler) validate(plan *api.Plan) error {
...
+		Status:   True,
+		Category: Critical,
+		Message:  "Target namespace is not valid.",
+	}
+	if plan.Spec.TargetNamespace == "" {
+		newCnd.Reason = NotSet
+		plan.Status.SetCondition(newCnd)
+		return
+	}
+	if len(k8svalidation.IsDNS1123Label(plan.Spec.TargetNamespace)) > 0 {
```
jortel, line 149: :+1:

PR #297, opened 2021-06-25
Fix updating VMs with batched validation result.

jortel:
Fix updating VMs with batched validation result.  When something failed or the revision of the validated VM no longer matched the _latest_ VM, we needed to continue processing other task results instead of returning.  As a result, subsequent task results were not updating their VMs and VMs would remain un-validated (in the DB) until the next schedule validation or an update triggered (re)validation of the VM.

This is a regression introduced late in 2.0.0 performance/scale fixes that also affects vSphere.


mansam (approved):
LGTM

PR #300, opened 2021-06-30
Added VM, vNIC and Disk properties needed for validation.

jortel:
Added VM, vNIC and Disk properties needed for validation.


mansam (approved):
LGTM

PR #303, opened 2021-07-05
Use ubi8-micro as base image. Update Golang to 1.15.13.

fdupont-redhat:
The `ubi8/ubi-micro` container image only weighs 39.1 MB, against 105 MB for the `ubi8/ubi-minimal` image. With this pull request, the total size of the controller image decreases from 156 MB to 89.9 MB. This will reduce download time and save space on the nodes where the controller runs.

At the same time, this pull request updates Golang to version 1.5.13, the latest available version.
I have built the image and tested it in a live cluster. No error, after a few tests.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,15 @@
 # Builder image
-FROM registry.access.redhat.com/ubi8/go-toolset:1.14.12 as builder
+FROM registry.access.redhat.com/ubi8/go-toolset:1.15.13 as builder
 ENV GOPATH=$APP_ROOT
-RUN env
 COPY . .
-RUN CGO_ENABLED=1 go build -o manager github.com/konveyor/forklift-controller/cmd/manager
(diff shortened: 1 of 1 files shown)


fdupont-redhat:
Self-approving.

mansam (approved):

Dockerfile
```diff
@@ -24,9 +26,3 @@ LABEL name="konveyor/forklift-controller" \
       io.openshift.tags="operator,konveyor,forklift,controller" \
       io.openshift.min-cpu="100m" \
       io.openshift.min-memory="350Mi"
-
-COPY --from=builder /opt/app-root/src/manager /usr/local/bin/manager
-
-RUN microdnf -y install tar && microdnf clean all
```
jortel, line 30: `Tar` is installed to support `oc cp` command needed to download memorycpu profiles and potentially the inventory sqlite3 DB for troubleshooting.  With this PR we can no longer do this.  The ubi-micro does not include a package manager so we'd need to switch back to ubi-minimal.

fdupont-redhat: I've created #326 to address that.

PR #304, opened 2021-07-07
Fix NAD watch and validation

jortel:
Fix NAD watch to match name only.
Add destination NAD ambiguity validation

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1971259


jortel:
Revising.

PR #316, opened 2021-07-19
Handle user add VM (asynchronously).

jortel:
Support both USER_ADD_VM(34) and USER_ADD_VM_STARTED(37) events.
There are two flows.
(A) With asynchronous tasks: First, USER_ADD_VM_STARTED is generated
(B) Without asynchronous tasks: First, USER_ADD_VM is generated

The disk and vNIC events follow.  Then, for asynchronous there a following:
USER_ADD_VM_FINISHED_SUCCESS or USER_ADD_VM_FINISHED_FAILURE (52).

We cannot simply rely on the USER_ADD_VM_FINISHED_SUCCESS create the VM because then, the update events for attaching disk and adding VMs will not find the VM in inventory to update (and fail).  So instead, the VM is created but then removed on USER_ADD_VM_FINISHED_FAILURE event.

https://bugzilla.redhat.com/show_bug.cgi?id=1982595

### DIFF
--- a/pkg/controller/provider/container/ovirt/reconciler.go
+++ b/pkg/controller/provider/container/ovirt/reconciler.go
@@ -405,13 +405,13 @@ func (r *Reconciler) refresh(ctx *Context) (err error) {
 			"event",
 			event)
 		var changeSet []Updater
-		changeSet, err = r.changeSet(ctx, event)
-		if err == nil {
-			err = r.apply(changeSet)
+		changeSet, applyErr := r.changeSet(ctx, event)
+		if applyErr == nil {
+			applyErr = r.apply(changeSet)
 		}
-		if err != nil {
+		if applyErr != nil {
 			r.log.Error(
-				err,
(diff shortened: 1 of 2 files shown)


pkg/controller/provider/container/ovirt/reconciler.go
```diff
@@ -405,13 +405,13 @@ func (r *Reconciler) refresh(ctx *Context) (err error) {
 			"event",
 			event)
 		var changeSet []Updater
-		changeSet, err = r.changeSet(ctx, event)
```
jortel, line 408: Using `err` within the loop will not result in a deterministic return value since we continue.

mansam (approved):
LGTM

pkg/controller/provider/container/ovirt/reconciler.go
mansam: Great catch

PR #318, opened 2021-07-23
Fix excessive oVirt validation.

jortel:
As part of periodic listing VMs that need to be validated, the latest policy version is requested from the policy-agent (validation service).  The version for oVirt and vSphere are different but the path used by the controller in the URL was hard coded for /vmware/.  As a result, when getting the version intended to be match for oVirt (latest=1), it was the policy version for vSphere (latest-5).  This caused the query of:
```
where revision != revisionValidated OR policyVersion != :version
```
to always match ALL oVirt VMs becuase `:version` was incorrect for oVirt.

The fix is to pass the _endpoint_ for retrieving the version from the validation service like is done for validation.

https://bugzilla.redhat.com/show_bug.cgi?id=1985480

---

Also:
- adjusted the logging a little based on experience troubleshooting.
- made the validation service CA optional to make testing easier.

### DIFF
--- a/pkg/controller/provider/container/vsphere/reconciler.go
+++ b/pkg/controller/provider/container/vsphere/reconciler.go
@@ -860,11 +860,6 @@ func (r Reconciler) applyEnter(tx *libmodel.Tx, u types.ObjectUpdate) error {
 		return liberr.Wrap(err)
 	}
 
-	r.log.V(3).Info(
-		"Model created.",
-		"model",
-		libmodel.Describe(m))
-
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


mansam (approved):
LGTM

pkg/controller/provider/container/vsphere/reconciler.go
```diff
@@ -860,11 +860,6 @@ func (r Reconciler) applyEnter(tx *libmodel.Tx, u types.ObjectUpdate) error {
 		return liberr.Wrap(err)
 	}
 
-	r.log.V(3).Info(
-		"Model created.",
-		"model",
-		libmodel.Describe(m))
-
```
jortel, line 867: Level=3 should not be used for per-object logging.
This will be logged by the _model_ layer at level=4.

PR #320, opened 2021-07-26
Fix processing RHV events for removal.

jortel:
Fix processing RHV events for removal of (dataCenter, cluster, host & VM)
Event handler using `Ref` but should be using `ID` which results  in _model.NotFound_ error and leaving the object in the inventory.  This is especially bad for VMs because it results in 500 errors trying to expand them in the web API.

Also, log validation task error as log.Error() to include more context.

https://bugzilla.redhat.com/show_bug.cgi?id=1986106


mansam (approved):
LGTM

PR #322, opened 2021-07-29
Fix post hook not running.

jortel:
The division of role/responsibility got blurred a little between:`Migration.updateVM()` and `Migration.updatePipeline()`.
In the process, updateVM() made an assumption that when the VMIO import completed, the VM status should be marked as completed.  As a result, the following `PostHook` step will never run.

I _think_ the fix is to move parts of updateVM() that interpreted VMIO Import status to updatePipeline() which only updates the step in the pipeline and anything that involves updating VM overall succeeded/failed status is moved to Migration.Run().


mansam (approved):
LGTM

PR #325, opened 2021-08-02
Inventory authorization required by default.

jortel:
Inventory authorization required by default.
Also, small modification to require a valid token for non-provider specific endpoints.


pkg/controller/provider/web/base/handler.go
```diff
@@ -112,7 +112,7 @@ func (h *Handler) setProvider(ctx *gin.Context) (status int) {
 		}
 		ctx.Header(ProviderHeader, uid)
 		h.Provider = h.Collector.Owner().(*api.Provider)
-		status = h.EnsureParity(h.Collector, time.Second*30)
+		status = h.EnsureParity(h.Collector, time.Second*10)
```
jortel, line 115: Reduced to eliminate _Gateway Timeout_ errors in openshift.

fdupont-redhat (approved):

PR #326, opened 2021-08-03
Switch back to ubi8/ubi-minimal base image

fdupont-redhat:
With #303 we removed `tar` from the image and this breaks the ability to copy files with `oc cp`. Since, `ubi8/ubi-micro` doesn't ship `tar` and doesn't allow to install it, we need to roll back to using `ubi8/ubi-minimal` and install `tar` during the image build.


jortel (approved):
LGTM

PR #327, opened 2021-08-04
Remove ImageConversion step from oVirt pipeline

mansam:
This PR causes the ImageConversion step to only be added to vSphere VM pipelines, and causes the ImportCreated phase to end when the DiskTransfer step is completed for VMs that don't have the ImageConversion step.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1980628


jortel (approved):
LGTM
Though after 2.1 I hope we can find a way to eliminate the need for conditional provider type logic for building the pipeline.

PR #328, opened 2021-08-04
Fix disks not detected and created when clone by template creation.

jortel:
Fix disk creation not detected when cloned by template creation.
When a template is created the reference VMs disks are cloned (created).  As a result, we have VMs created with the template (thin provisioned) that references disks that are not in inventory. We need to refresh the disks when a template is created.  
It may seem _heavy-handed_ to fetch all of the disks but it's simpler than fetching the template and inspecting the disks.  Also, disk resources are relatively small resources and I doubt that template creation is a high volume event.

--- 
This PR also improves some logging.

---

This PR also address the latency between when events are generated and the current state of the RHV model.
For example, here is a list of events grabbed in a query:
- VM Created
- Disk attached.
- Snapshot created.
- VM started.
- VM deleted.  <--- note.

When all of these are processed, the VM will no longer be in the RHV db.  So, we need to handle this gracefully.

### DIFF
--- a/pkg/controller/provider/container/ovirt/model.go
+++ b/pkg/controller/provider/container/ovirt/model.go
@@ -2,6 +2,7 @@ package ovirt
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"github.com/go-logr/logr"
 	liberr "github.com/konveyor/controller/pkg/error"
@@ -43,8 +44,7 @@ const (
 	USER_REMOVE_HOST = 44
 	// VM
 	USER_ADD_VM                           = 34
-	USER_ADD_VM_STARTED                   = 37
-	USER_ADD_VM_FINISHED_FAILED           = 52
+	USER_ADD_VM_FINISHED_SUCCESS          = 53
 	USER_UPDATE_VM                        = 35
 	USER_REMOVE_VM                        = 113
 	USER_ADD_DISK_TO_VM_SUCCESS           = 97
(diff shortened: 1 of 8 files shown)


pkg/controller/provider/container/ovirt/model.go
```diff
@@ -41,10 +42,12 @@ const (
...
 	USER_UPDATE_HOST = 43
 	USER_REMOVE_HOST = 44
+	// Template
+	USER_ADD_VM_TEMPLATE                  = 48
+	USER_ADD_VM_TEMPLATE_FINISHED_SUCCESS = 51
 	// VM
 	USER_ADD_VM                           = 34
-	USER_ADD_VM_STARTED                   = 37
-	USER_ADD_VM_FINISHED_FAILED           = 52
+	USER_ADD_VM_FINISHED_SUCCESS          = 53
```
jortel, line 50: USER_ADD_VM_STARTED not useful.
USER_ADD_VM generated for synchronous create.
USER_ADD_VM_TEMPLATE_FINISHED_SUCCESS generated when created asynchronous task succeeds.

Since we are gracefully handling (ignoring) events between USER_ADD_VM_STARTED and USER_ADD_VM_TEMPLATE_FINISHED_SUCCESS we can just trigger off of USER_ADD_VM_TEMPLATE_FINISHED_SUCCESS.

mansam (approved):
LGTM

PR #334, opened 2021-08-09
Fix vSphere traversal.

jortel:
Fix vSphere traversal to support:
- stand-alone host
- host nested under cluster sub-folder.

### DIFF
--- a/pkg/controller/provider/web/base/auth.go
+++ b/pkg/controller/provider/web/base/auth.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"github.com/gin-gonic/gin"
 	liberr "github.com/konveyor/controller/pkg/error"
-	"github.com/konveyor/controller/pkg/ref"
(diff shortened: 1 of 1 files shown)


pkg/controller/provider/container/vsphere/collector.go
```diff
@@ -34,15 +34,16 @@ const (
...
-	Cluster        = "ClusterComputeResource"
-	Host           = "HostSystem"
-	Network        = "Network"
-	DVPortGroup    = "DistributedVirtualPortgroup"
-	DVSwitch       = "VmwareDistributedVirtualSwitch"
-	Datastore      = "Datastore"
+	Folder          = "Folder"
+	VirtualMachine  = "VirtualMachine"
+	Datacenter      = "Datacenter"
+	ComputeResource = "ComputeResource"
```
jortel, line 40: Add: _ComputeResource_ used in traversal spec.

PR #335, opened 2021-08-09
Fix support for standalone host.

jortel:
Fix support for stand-alone Hosts.

A host system can be contained by any `ComputeResource`.  A cluster IsA  `ComputeResource`.  Standalone host will be contained by a `ComputeResource` (that is not a Cluster).

Both `ComputeResource` and `ClusterComputeResources` are stored in the model as `Cluster`.  However, the base `ComputeResource` containing stand-alone hosts are denoted using the `variant=ComputeResource` field.


mansam (approved):
LGTM

PR #339, opened 2021-08-16
Removed Host.VMs property.

jortel:
The `Host.VMs` list is very large and is updated whenever a VM is created,deleted.  As a result, when a VM is added and the list is updated, the controller re-validates all of the VMs on the host (because the host changed).
The `VM.Host` property (back-ref) provides that same linkage.

### DIFF
--- a/pkg/controller/provider/container/vsphere/watch.go
+++ b/pkg/controller/provider/container/vsphere/watch.go
@@ -470,16 +470,22 @@ func (r *HostEventHandler) validate(host *model.Host) {
 	defer func() {
 		_ = tx.End()
 	}()
-	for _, ref := range host.Vms {
-		vm := &model.VM{}
-		vm.WithRef(ref)
-		err = tx.Get(vm)
-		if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/controller/provider/container/vsphere/watch.go
```diff
@@ -470,16 +470,19 @@ func (r *HostEventHandler) validate(host *model.Host) {
...
-	for _, ref := range host.Vms {
-		vm := &model.VM{}
-		vm.WithRef(ref)
-		err = tx.Get(vm)
-		if err != nil {
-			r.log.Error(err, "VM (get) failed.")
-			return
-		}
+	list := []model.VM{}
+	err = tx.List(
```
mansam, line 474: Is this list restricted to just the VMs with references back to the host? It looks like it's querying all of them.

jortel: Fixed.

mansam (approved):

PR #344, opened 2021-08-24
Improved REST Resource.Path building.

jortel:
Improved REST Resource.Path building.  The primary driver is to support UI disambiguating networks re: https://github.com/konveyor/forklift-ui/pull/750

Shift responsibility for path building to the `web` package.  This is really a web layer concern. Add a `PathBuilder` to both vSphere and oVirt packages and use _everywhere_.  Uses caching to reduce DB burden.  Only _parents_ are cached so VMs will never be cached which means a relatively small write-few/read-many cache. The path builder logs errors instead of returning them.

Ensure endpoint handler non-exported methods to have: a pointer receiver.  Avoid copying the handler unnecessarily. This is safe (unlike the exported Get() and List() methods).

Add _defer_ for error handling to reduce duplication.


mansam (approved):
LGTM

PR #346, opened 2021-09-13
vSphere VM client and Client interface

mansam:
* Adds vSphere VM client
* Adds Client interface
* Adds Client constructors for oVirt and vSphere.


jortel (approved):
LGTM

PR #350, opened 2021-09-15
Add support for single VM import to oVirt builder

mansam:
This PR expands the `Builder` interface with four methods that will enable building DataVolumes, VirtualMachines, and ConfigMaps respectively, and implements the oVirt versions of those methods. `Builder.Import` will be removed in a subsequent PR, but is left for the time being to avoid having to deal with the interface changes everywhere at once.


jortel (approved):
LGTM

PR #355, opened 2021-09-20
Single VM import: kubevirt, itineraries, cleanup

mansam:
* Add support for single VM import to kubevirt.go
  (creating VMs and Datavolumes, guest conversion, etc)
* Introduce new cold and warm migration itineraries to
  migration.go
* Create DVs by specifying Storage rather than PVC.
* Clean up remains of VMIO integration.


jortel (approved):
LGTM

PR #356, opened 2021-09-22
Bug 1942651: vSphere: pass changeID instead of snapshot to CDI

mansam:
This adds support for the improved incremental backup process for VDDK in CDI, along with a feature gate to enable it. This feature allows vSphere snapshots to be cleaned up after each precopy, rather than having to wait to clean up at the end. In addition, this causes the snapshots to be removed when the plan is archived or a VM migration is canceled.


jortel (approved):
LGTM

PR #358, opened 2021-09-24
Graphql endpoint and queries for VSphere and OVirt providers

gildub:
Add GraphQL feature, which is activated only if environment variable `GRAPHQL_ENABLED` is enabled (for example GRAPHQL_ENABLE=true)

With following endpoinds:
- `/graphql` used by API clients 
- `/graphql/playground` built-in client (See screenshot below) used only in development. (Need to add a debug flag)

Offers queries to traverse VSphere and OVirt providers all the way down to VMs, through datacenters, clusters, hosts, networks, datastores, disks.

Screenshot from 2021-10-08 18-23-12

https://github.com/konveyor/forklift-controller/issues/359


pkg/controller/provider/web/graph/model/models_gen.go
```diff
@@ -0,0 +1,10 @@
+// Code generated by github.com/99designs/gqlgen, DO NOT EDIT.
+
+package model
+
+type VsphereProvider struct {
+	ID      string `json:"id"`
+	Name    string `json:"name"`
+	Kind    string `json:"kind"`
+	Product string `json:"product"`
```
jortel, line 9: What is a product?

gildub: That's a reminiscent of an VMware object in Opal schema. 
Never mind, I'll remove it.

jortel:
Why is the generator a dependency?
Seems the generator would be used initially to generate the code based on the yaml file.  And, potentially by the makefile with could get it following the existing pattern of `controller-gen`.

See: https://github.com/konveyor/forklift-controller/issues/359#issuecomment-927919363

gildub:
@jortel, sure but where do you see it defined as a dependency? I might be missing something meanwhile I didn't add the generator to the modules.

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/konveyor/forklift-controller
 go 1.14
 
 require (
+	github.com/99designs/gqlgen v0.14.0
```
jortel, line 6: This isn't a code generator?

gildub: Oh I see, that's actually both, the GraphQL library which uses a generator.

pkg/controller/provider/web/base/handler.go
```diff
@@ -1,21 +1,23 @@
...
-	"net/http"
-	"strconv"
-	"strings"
-	"time"
 )
 
 //
 // Root - all routes.
 const (
+	GraphqlRoot   = "graphql"
```
jortel, line 20: This should go in the GraphHander .go file.

pkg/controller/provider/web/graph/doc.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"github.com/konveyor/controller/pkg/inventory/container"
+	libweb "github.com/konveyor/controller/pkg/inventory/web"
+	"github.com/konveyor/forklift-controller/pkg/controller/provider/web/base"
+)
+
+//
+// Routes
+const (
+	Root = base.GraphqlRoot
```
jortel, line 12: Let's just define the Root = "graphql" here.

pkg/controller/provider/web/graph/graphql.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/99designs/gqlgen/graphql/handler"
+	"github.com/99designs/gqlgen/graphql/playground"
+)
+
+var log = logging.WithName("web|graphql")
+
+//
+// Routes.
+const (
+	GraphqlRoot = Root
```
jortel, line 18: Let's just define the GraphRoot = "graphql" here.

pkg/controller/provider/web/graph/graphql.go
```diff
@@ -0,0 +1,52 @@
...
+	base.Handler
+}
+
+//
+// Add routes to the `gin` router.
+func (h *GraphHandler) AddRoutes(e *gin.Engine) {
+	e.POST(GraphqlRoot, h.GraphqlHandler())
+	e.GET(GraphqlRoot+"/playground", playgroundHandler())
+}
+
```
jortel, line 33: To be consistent with other handlers, recommend something like:
```
e.POST(GraphqlRoot, h.Post)
e.GET(GraphqlRoot+"/playground", h.Get)

func (h GraphHandler) Post(ctx *gin.Context) {
    err := r.updateContext(ctx)
    if err != nil {
        ctx.Status(http.StatusInternalServerError)
        return
    }

    r.qqlh.ServeHTTP(ctx.Writer, ctx.Request)
}

func (h GraphHandler) Get(ctx *gin.Context) {
    err := r.updateContext(ctx)
    if err != nil {
        ctx.Status(http.StatusInternalServerError)
        return
    }

    r.playground.ServeHTTP(ctx.Writer, ctx.Request)
}

func (h *GraphHandler) updateContext(ctx *gin.Context) (err error) {
    # inject container into the ctx.Request.Context
    return
}
```

The GraphHandler.gqlh and GraphHandler.playground can be assigned when the Handler is constructed in doc.go.

gildub: Makes sense.

pkg/controller/provider/web/graph/doc.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+//
+// Build all handlers.
+func Handlers(container *container.Container) []libweb.RequestHandler {
+	return []libweb.RequestHandler{
+		&GraphHandler{
+			Handler: base.Handler{
+				Container: container,
```
jortel, line 21: :+1:

pkg/controller/provider/web/graph/graphql.go
```diff
@@ -0,0 +1,57 @@
...
+//
+// GraphQL Handler.
+func (h GraphHandler) Post() gin.HandlerFunc {
+	handler := handler.NewDefaultServer(generated.NewExecutableSchema(generated.Config{Resolvers: &Resolver{}}))
+
+	return func(c *gin.Context) {
+		ctx := context.WithValue(c.Request.Context(), "HandlerContainer", h.Container)
+		c.Request = c.Request.WithContext(ctx)
+		c.Next()
```
jortel, line 44: I don't think you want to do c.Next() here.  I think c.Next() will advance to the next handler.

pkg/controller/provider/web/graph/graphql.go
```diff
@@ -0,0 +1,57 @@
...
+//
+// Add routes to the `gin` router.
+func (h *GraphHandler) AddRoutes(e *gin.Engine) {
+	e.POST(GraphqlRoot, h.Post())
+	e.GET(GraphqlRoot+"/playground", h.Get())
+}
+
+//
+// GraphQL Handler.
+func (h GraphHandler) Post() gin.HandlerFunc {
```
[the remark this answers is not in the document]
gildub: Somehow I can think with function first class citizen only.

go.mod
gildub: @jortel, gqlgen is used here: https://github.com/konveyor/forklift-controller/pull/358/files#diff-72eef241725fe361fe8d4e1e9bbd25b25915aa2ae9085c80175da93856bcdbb6R15-R16

jortel:
I really appreciate the work that went into this PR. However, it adds an additional API for which there is no known need / use case.
It is already out of sync with the REST API and should it be merged will require on-going maintenance.  Closing for now.  It can be re-opened should the need arise to support a GraphQL API.

PR #361, opened 2021-09-30
Archive plans

mansam:
Add support for archiving plans.

* Resources are now left behind when a migration fails. These resources are cleaned up when the Plan is retried, or when the Plan is archived. (Resources from canceled VMs are not retained)
* Setting the `archived` flag on the Plan causes all remaining resources (failed VMs, conversion and importer pods, configmaps, secrets, etc) to be cleaned up and the plan to get an `Archived` condition.
* kubevirt.go `Delete` methods now delete all left over resources for a plan rather than just those from a specific migration.

Misc

* Remove obsolete `RequiresConfigMap()` from Builder interface.

### DIFF
--- a/pkg/controller/plan/controller.go
+++ b/pkg/controller/plan/controller.go
@@ -360,7 +360,6 @@ func (r *Reconciler) execute(plan *api.Plan) (reQ time.Duration, err error) {
 	if err != nil {
 		return
 	}
-
 	//
 	// Find pending migrations.
 	pending := []*api.Migration{}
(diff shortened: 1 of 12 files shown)


pkg/controller/plan/controller.go
```diff
@@ -254,6 +262,31 @@ func (r Reconciler) Reconcile(request reconcile.Request) (result reconcile.Resul
 	return
 }
 
+//
+// Archive the plan.
+func (r *Reconciler) archive(plan *api.Plan) (err error) {
```
jortel, line 267: :+1:

PR #366, opened 2021-10-13
Fix pointer bug in ovirt builder's mapDisks

mansam:
https://bugzilla.redhat.com/show_bug.cgi?id=2012130


pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -294,7 +294,8 @@ func (r *Builder) mapFirmware(vm *model.Workload, cluster *model.Cluster, object
 
 func (r *Builder) mapDisks(vm *model.Workload, dataVolumes []cdi.DataVolume, object *cnv.VirtualMachineSpec) {
 	dvMap := make(map[string]*cdi.DataVolume)
-	for _, dv := range dataVolumes {
+	for i := range dataVolumes {
+		dv := dataVolumes[i]
 		dvMap[dv.Spec.Source.Imageio.DiskID] = &dv
 	}
```
jortel, line 300: Recommend:
```
for i := range dataVolumes {
  dv := &dataVolumes[i]
  dvMap[dv.Spec.Source.Imageio.DiskID] = dv
}
```
to eliminate the shallow copy of the _DataVolume_ unless it's intended.

PR #373, opened 2021-10-18
Fix destination VM power on condition

mansam:
Rather than always powering on the destination for warm migrations, defer to the power state of the source VM like cold migrations do.


jortel (approved):
LGTM

PR #384, opened 2021-11-02
Bug 2015940: Omit Migration label from List call when building the VM CR map

mansam:
Removing the Migration label from the label selector in `ListVMs` allows the VM CR map to be built when there is not an active migration.


jortel (approved):
LGTM

PR #387, opened 2021-11-04
Bug 2018974: Add finalizer to archive Plan when it is deleted

mansam:


pkg/controller/plan/controller.go
```diff
@@ -229,9 +246,28 @@ func (r Reconciler) Reconcile(request reconcile.Request) (result reconcile.Resul
...
+		if err != nil {
+			return
+		}
+		return
+	}
+
+	// Archive the plan due to a user request.
 	if plan.Spec.Archived {
 		r.archive(plan)
+		plan.Status.SetCondition(
```
mansam, line 263: Setting the condition now happens outside of `archive()` and only when the archive is user requested rather than prompted by deletion. This is because we can't do anything except remove finalizers once the deletion timestamp is set.

pkg/controller/plan/controller.go
```diff
@@ -204,8 +210,19 @@ func (r Reconciler) Reconcile(request reconcile.Request) (result reconcile.Resul
...
 	}()
 
-	// Don't reconcile if the plan is archived.
-	if plan.Spec.Archived && plan.Status.HasCondition(Archived) {
+	// Set archive finalizer to ensure resources are
+	// cleaned up if the plan is deleted without
+	// having been manually archived.
+	if plan.DeletionTimestamp.IsZero() && !plan.HasFinalizer(ArchiveFinalizer) {
+		plan.AddFinalizer(ArchiveFinalizer)
+		err = r.Update(context.TODO(), plan)
```
mansam, line 218: @jortel do you think there would be any issue with changing the Plan status subresource `Update()` call at the end of `Reconcile()` to just update the whole plan? That would allow me to leave out the `Update()` call here.

jortel (changes requested):
I **strongly** recommend against using _finalizers_.  Every attempt to use them for this kind of clean up has cause a lot of problems in the past.  Especially as it relates to deleting the entire MTV namespace.

PR #389, opened 2021-11-15
Bug 2012564: VMStatus.RestorePowerState is not a required field

mansam:
https://bugzilla.redhat.com/show_bug.cgi?id=2012564


fdupont-redhat (approved):
LGTM

PR #390, opened 2021-11-16
Use "Start" and "End" in the Precopy status

mansam:
This effectively reverts the change to have `Precopy` contain a `Timing`, due to it causing more trouble that it was worth for upgrades from 2.1 to 2.2. In the future we can revisit this with some additional work in the controller to handle the data migration.


jortel (approved):
LGTM

PR #392, opened 2021-11-29
Bug 2024138: Log rather than fail if importer pod isn't found

mansam:
If the CDI importer pod for a datavolume isn't found while trying to update the status of the disk transfer, then we should just log and continue rather than causing the migration to fail. CDI recreates the importer pod if it's removed while the import is in progress, so we can expect the importer pod not being found to be a transient issue.


fabiendupont:
Does it also fix the nil pointer exception?

jortel (approved):
LGTM

PR #393, opened 2021-11-29
Bug 2027336: VM validation event tuning and logging.

fabiendupont:
Using _transaction labels_ introduced in controller/v0.6.1, **ignore** _echo_ events generated when VMs are updated after they are validated.  Also, use buffered _input_ channel to add some _"slack"_.  Both prevent the VM model event handler from getting blocked on the first event during initial validation on provider startup.  For very large vCenter providers (node-5), I have seen watch queue overflow log entries:

"full queue, event discarded."

Log the "VM validated" at level: 3 which matches the default logging level when deployed by the operator.

Signed-off-by: Fabien Dupont <fabiendupont@pm.me>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/gin-gonic/gin v1.7.2
 	github.com/go-logr/logr v0.3.0
 	github.com/k8snetworkplumbingwg/network-attachment-definition-client v1.1.0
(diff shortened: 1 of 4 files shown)


fabiendupont:
Wrong branch... 🤦🏼 
Closing

PR #400, opened 2021-12-15
Let the StorageProfile decide volume/access modes

mansam:
Stops setting the default volume and access modes in the datavolume builder, and instead only sets them if they were specified in the StorageMap. In all other cases the StorageProfile for the selected StorageClass will be used by CDI to determine the defaults.
Also removes some code related to Provisioners from the builders. The Provisioner API will still need to be removed in the future.


fabiendupont (approved):
LGTM

PR #402, opened 2022-01-05
Event latched validation.

jortel:
Revised the validation algorithm to be both timed and event latched instead of timed and event driven.
This approach is:
- simpler
- smaller memory footprint.
- most important: eliminate the possibility of blocking the VMEventHandler and overflowing[1] the watch queue. 

The queue overflow does not break anything but causes the customer concern and could send support/eng on a wild goose chase.

Note: This requires a feature in konveyor/controller to support update with predicate (not yet released).

---

[1] {"level":"info","ts":1637843998.908607,"logger":"collector|vsphere","msg":"full queue, event discarded","provider":"openshift-mtv/vmware","error":"full queue, event discarded"}


mansam (approved):
LGTM

mansam (approved):
LGTM

PR #405, opened 2022-01-17
Bug 2036482: Retrieve PVCs when listing VMs and DVs

mansam:
With the change to using StorageProfiles, the DV specification no longer is guaranteed to contain a value for VolumeMode or AccessModes. We need to look at the created PVC to discover which modes were used, so that the libvirt domain can be created correctly.

https://bugzilla.redhat.com/show_bug.cgi?id=2036482


jortel (approved):
LGTM

PR #406, opened 2022-01-20
Fix OpenShift VM template processing

mansam:
* Don't overwrite the template spec if it exists.
* Do overwrite existing disks and volumes.
* Cause vSphere GuestID not to be wiped out
  of inventory when the VM is powered off.


jortel (approved):
LGTM

PR #408, opened 2022-01-25
Add Hook Documentation

jmontleon:


docs/hooks.md
```diff
@@ -0,0 +1,242 @@
...
+    - id: vm-2861
+      hooks:
+        - hook:
+            namespace: konveyor-forklift
+            name: playbook
+          step: PreHook
+...
+```
+
+# Adding a MigHook CR
```
jortel, line 30: Please remove the `Mig` prefix.  MTV does not prefix everything with `Mig`.

docs/hooks.md
```diff
@@ -0,0 +1,242 @@
+# Introduction
+Migration hooks provide a means for running custom code at points in the migration to handle logic unique to a migration. This may include writing new configuration files or installing additional packages in preparation for migration to prepare for the new environment.
+
+# Default hook image
+The default hook image is `quay.io/konveyor/hook-runner:latest`. It is based off of the Ansible Runner image with the addition of python-openshift to provided Ansible k8s resources as well as a recent oc binary.
+
+# Hook execution
+When an Ansible playbook is provided as part of a migration hook it will be mounted into the hook container as a ConfigMap. In either case the hook container will be run as job on the desired cluster, using the default ServiceAccount in the konveyor-forklift namespace.
```
jortel, line 8: Or, the user may specify a _serviceAccount_ in the Hook specification.

docs/hooks.md
```diff
@@ -0,0 +1,242 @@
...
+
+```
+
+# Storing additional information in secrets
+If you wish to access additional information stored in secrets from within the hook container you may, but you will need to provide additional access to the default account in the konveyor-forklift namespace. An example role and rolebinding to provide acecss to secrets is included here.
```
jortel, line 61: I will tweak your example to set the Hook SA to the controller SA but we should word this to encourage users to do this instead of granting read on secret to the default SA.

docs/hooks.md
jortel: We should probably recommend they set the SA on the Hook to `serviceAccount: forklift-controller`.

docs/hooks.md
jmontleon: I removed the role(binding) stuff, added the SA to the example, and mentioned the possibility of changing it.

jortel (approved):
LGTM.  Thanks @jmontleon.

PR #410, opened 2022-01-27
Bug 2036208: Fix handling of multiple nics attached to same network

mansam:
* Adds a validation to ensure that no more than one
  nic is mapped to the pod network at a time.
* Create kubevirt nics for all source nics attached
  to the same network.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2036208


jortel (approved):
LGTM

PR #414, opened 2022-02-09
Bug 2048505: oVirt: Use disk snapshot IDs in DV checkpoints

mansam:
https://bugzilla.redhat.com/show_bug.cgi?id=2048505


jortel (approved):
LGTM

PR #418, opened 2022-02-17
Make sure snapshot disk ID matches target disk ID.

mrnold:
Fix a bug with multi-disk RHV warm imports, where every datavolume would be given the same disk snapshot ID for different disks (CDI has the same bug and accepts the import).

Found while investigating [RHBZ#2055201](https://bugzilla.redhat.com/show_bug.cgi?id=2055201), but not actually a complete fix.


mansam (approved):
LGTM!

jortel:
back-ported to 2.3.0.

PR #429, opened 2022-05-26
Bug 2087916: Request oVirt DataVolume of ActualSize if larger than ProvisionedSize. 

mrnold:
For [2087916](https://bugzilla.redhat.com/show_bug.cgi?id=2087916), choose the larger of ActualSize or ProvisionedSize so that CDI does not run out of space trying to download the snapshot. There are some cases where ActualSize can be larger.


jortel (approved):
LGTM

PR #434, opened 2022-06-28
NO ISSUE: Modifying Makefile

bkhizgiy:
Modifying Makefile, with current implementation for newer go version controller-gen is not being installed properly and the build failed, changing "go get" to "go install" and removing the temp dir since it's not required, the binary file is placed under $HOME/go/bin. 
Also changing controller-gen to a newer version (still support legacy features) and switching docker with podman.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,12 +68,7 @@ docker-push:
 controller-gen:
 ifeq (, $(shell which controller-gen))
 	@{ \
-	set -e ;\
-	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
-	cd $$CONTROLLER_GEN_TMP_DIR ;\
-	go mod init tmp ;\
-	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.5.0 ;\
-	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
+	go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.6.2 ;\
 	}
 CONTROLLER_GEN=$(GOBIN)/controller-gen
 else


Makefile
```diff
@@ -72,7 +72,7 @@ ifeq (, $(shell which controller-gen))
 	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
 	cd $$CONTROLLER_GEN_TMP_DIR ;\
 	go mod init tmp ;\
-	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.5.0 ;\
+	go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.5.0 ;\
```
ahadas, line 75: can we drop the line above then?

michalskrivanek: also 0.5.0 doesn't work for me on  go1.18.3 darwin/arm64, while 0.6.2 does and it seems to still support the needed legacy features.

michalskrivanek:
it probably needs to run with older versions too so it may need a bit more work

Makefile
```diff
@@ -55,25 +55,20 @@ generate: controller-gen
 # Build the docker image
 #docker-build: test
 docker-build:
-	docker build . -t ${IMG}
+	podman build . -t ${IMG}
 	@echo "updating kustomize image patch file for manager resource"
 	sed -i'' -e 's@image: .*@image: '"${IMG}"'@' ./config/default/manager_image_patch.yaml
 
 # Push the docker image
```
liranr23, line 62: worth renaming all docker (comments, targets) to podman?

jortel: Personally, I don't find this target useful anyway.  I just build/push the image with one command.

ahadas: > Personally, I don't find this target useful anyway. I just build/push the image with one command.

+1
btw, where can I find that manager_image_patch.yaml?

@bkhizgiy how about dropping all the docker to podman changes from this PR? we were distracted and this PR had value also without those changes

bkhizgiy:
@jortel @mansam Hey, can you please take a look when you have the time, it's a really minor change but required for a successful build.

mansam (approved):

PR #435, opened 2022-06-29
[WIP] Enable minikube: Add ignore of empty Authorization

mnecas:
Related to the https://github.com/konveyor/forklift-ui/pull/969
The UI patch allows to disable the login but in the controller, it is still needed.
This patch allows the requests without `Authorization` as in minikube there is no oauth.


jortel:
Disabling auth is already supported by the controller.  See [here](https://github.com/konveyor/forklift-controller/blob/ce9361067d41b4754f4a9ce3e73446c29bc5868e/pkg/controller/provider/web/base/handler.go#L157).
The environment variable `AUTH_REQUIRED` needs to be set _false_ by the operator when installing in minikube.

mnecas:
Thanks for showing it to me! Closing

PR #436, opened 2022-07-07
use access token with raw http client

bennyz:
This would make the client usable with setups that have keycloak enabled, and is the standard way to access the API.

Since we cannot rely on the "exp" field return from oVirt, the current approach refreshes the client every 30 minutes (default session timeout) by requesting a new token or when the token is expired (in the event it doesn't use the default Long.MAX_VALUE).

### DIFF
--- a/pkg/controller/provider/container/ovirt/client.go
+++ b/pkg/controller/provider/container/ovirt/client.go
@@ -112,7 +112,17 @@ func (r *Client) connect() (err error) {
 
 	response := &ovirtTokenResponse{}
 	url.RawQuery = values.Encode()
-	client.Get(url.String(), response)
+	status, err := client.Get(url.String(), response)
+	if err != nil {
+		return
+	}
+
+	// Providing bad credentials when requesting the token results
+	// in 400, and not 401. So checking for != 200 instead
+	if status != http.StatusOK {
+		err = liberr.New("Request for token failed", "status", status)
+		return
+	}
 
 	// Set the access token we received
 	client.Header = http.Header{
--- a/pkg/controller/provider/container/ovirt/client.go
+++ b/pkg/controller/provider/container/ovirt/client.go
@@ -3,7 +3,6 @@ package ovirt
 import (
 	"crypto/tls"
 	"crypto/x509"
-	"encoding/base64"
 	"net"
 	"net/http"
 	liburl "net/url"
@@ -34,6 +33,10 @@ type Client struct {
 	client *libweb.Client
 	// Secret.
 	secret *core.Secret
+}
+
+type ovirtTokenResponse struct {
+	AccessToken string `json:"access_token"`
 }
 
 //
@@ -73,12 +76,32 @@ func (r *Client) connect() (err error) {
 			TLSClientConfig:       TLSClientConfig,
 		},
 	}
+
+	url, err := liburl.Parse(r.url)
+	if err != nil {
+		return
+	}
+
+	url.Path = "/ovirt-engine/sso/oauth/token"
+	values := url.Query()
+	values.Add("grant_type", "password")
+	values.Add("username", string(r.secret.Data["user"]))
+	values.Add("password", string(r.secret.Data["password"]))
+	values.Add("scope", "ovirt-app-api")
+
 	client.Header = http.Header{
 		"Accept": []string{"application/json"},
-		"Authorization": []string{
-			"Basic",
-			r.auth()},
-		"Version": []string{"4"},
+	}
+
+	response := &ovirtTokenResponse{}
+	url.RawQuery = values.Encode()
+	client.Get(url.String(), response)
+
+	// Set the access token we received
+	client.Header = http.Header{
+		"Accept":        []string{"application/json"},
+		"Authorization": []string{"Bearer " + response.AccessToken},
+		"Version":       []string{"4"},
 	}
 
 	r.client = client
@@ -132,21 +155,6 @@ func (r *Client) get(path string, object interface{}, param ...libweb.Param) (er
 	default:
 		err = liberr.New(http.StatusText(status))
 	}
-
-	return
-}
-
-//
-// Basic authorization user.
-func (r *Client) auth() (user string) {
-	user = strings.Join(
-		[]string{
-			string(r.secret.Data["user"]),
-			string(r.secret.Data["password"]),
-		},
-		":")
-
-	user = base64.StdEncoding.EncodeToString([]byte(user))
 
 	return
 }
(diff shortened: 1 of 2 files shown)


ahadas (approved):
as discussed offline - this method is supported on ovirt 4.3 so that's ok from backward compatibility perspective. without it, forklift is not able to access new deployments of ovirt 4.5

cc @jortel

bennyz:
> > as discussed offline - this method is supported on ovirt 4.3 so that's ok in terms of backward compatibility. without this, forklift is not able to access new deployments of ovirt 4.5
> > cc @jortel
> 
> Does the token expire?

Where can I check for expiry and reconnect if needed? Is system(), in:
https://github.com/konveyor/forklift-controller/blob/7c95747942453b50b9a130097a88af4673719e0b/pkg/controller/provider/container/ovirt/client.go#L147-L163
Expected to be called periodically?

pkg/controller/provider/container/ovirt/collector.go
```diff
@@ -69,10 +70,18 @@ func New(db libmodel.DB, provider *api.Provider, secret *core.Secret) (r *Collec
...
+			provider.GetNamespace(),
+			provider.GetName()))
 	r = &Collector{
 		client: &Client{
-			url:    provider.Spec.URL,
-			secret: secret,
+			url:           provider.Spec.URL,
+			secret:        secret,
+			log:           clientLog,
+			clientTimeout: time.Minute * 30, // TODO: hardcoded for now, move to provider spec
```
bennyz, line 83: Made it hardcoded to get approval on the direction, if this is accepted it should be configurable in the provider spec

mansam (approved):
LGTM, but @jortel should also take a look.

mnecas (approved):
Just tested with 4.5 with keycloak and 4.4 without and works perfectly.

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -64,15 +86,43 @@ func (r *Client) connect() (err error) {
...
 
 	r.client = client
+	r.created = time.Now()
+
+	expiration, err := strconv.ParseInt(response.Expiration, 10, 64)
+	if err != nil {
+		return
+	}
+
+	r.expiration = time.Now().Local().Add(time.Duration(expiration))
```
mnecas, line 125: If the user does not enter correct auth credentials it will fail with...
```
{"level":"info","ts":1660222302.1400778,"logger":"provider","msg":"Connection test failed.","reason":"strconv.ParseInt: parsing \"\": invalid syntax"}
```
Because the `response.Expiration` and `response.AccessToken` are empty strings.
The `msg` is correct but it would be better to have some more proper `reason`.

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -31,12 +33,32 @@ type Client struct {
...
+
+type ovirtTokenResponse struct {
+	AccessToken string `json:"access_token"`
+	Expiration  string `json:"exp"`
 }
 
 //
 // Connect.
 func (r *Client) connect() (err error) {
+	if time.Now().After(r.created.Local().Add(r.clientTimeout)) {
```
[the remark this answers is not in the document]
bennyz: From what I chcked, yes:
https://github.com/konveyor/forklift-controller/blob/8d67009dee1bfe152213a9df1948f1f59449fe69/pkg/controller/provider/container/ovirt/collector.go

pkg/controller/provider/container/ovirt/client.go
bennyz: added error check above

yaacov (approved):
works for me :fire:

ahadas:
@mansam hi, this can help us internally, what blocks this from getting in?

pkg/controller/provider/container/ovirt/collector.go
mansam, line 83: Please fix this now rather than leaving a `TODO`. The Provider spec contains a `settings` map for provider-type-specific fields, which could be used for this, with a default of 30 minutes if a value is not set.

mansam (changes requested):
Looks fine other than the TODO. @jortel would you mind taking a look? This is an inventory change.

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -73,15 +93,53 @@ func (r *Client) connect() (err error) {
...
+		"Authorization": []string{"Bearer " + response.AccessToken},
+		"Version":       []string{"4"},
 	}
 
 	r.client = client
+	r.created = time.Now()
+
+	expiration, err := strconv.ParseInt(response.Expiration, 10, 64)
+	if err != nil {
+		return
```
ahadas, line 139: looking at the error Martin mentioned below, if we return that error back I'm afraid it won't be so clear to users - how about returning an error with better description, i.e., saying that we failed to parse the expiration part

pkg/controller/provider/container/ovirt/collector.go
```diff
@@ -69,10 +73,28 @@ func New(db libmodel.DB, provider *api.Provider, secret *core.Secret) (r *Collec
...
+		"provider",
+		libpath.Join(
+			provider.GetNamespace(),
+			provider.GetName()))
+
+	var err error
+	clientTimeout := DefaultClientTimeout
+	if timeout, ok := provider.Spec.Settings["ovirtClientTimeout"]; ok {
+		if clientTimeout, err = time.ParseDuration(timeout); err != nil {
+			log.Error(err, "Couldn't parse timeout, falling back to default")
```
ahadas, line 86:
```suggestion
			log.Error(err, "Couldn't parse http client timeout, falling back to default")
```

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -33,14 +33,34 @@ type Client struct {
 	// Raw client.
 	client *libweb.Client
 	// Secret.
-	secret *core.Secret
+	secret        *core.Secret
+	created       time.Time
+	clientTimeout time.Duration
+	expiration    time.Time
```
ahadas, line 39: how about
```suggestion
	accessTokenExpiration    time.Time
```

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -33,14 +33,34 @@ type Client struct {
 	// Raw client.
 	client *libweb.Client
 	// Secret.
-	secret *core.Secret
+	secret        *core.Secret
+	created       time.Time
+	clientTimeout time.Duration
```
ahadas, line 38: I don't see a need to store both, we can have just a single field for 'client expiration', no?

ahadas (commented):
looks good and many of us use it in order to be able to work with latest ovirt - how about moving this PR to github.com/kubev2v/forklift?

ahadas:
let's post to kubev2v/forklift

PR #437, opened 2022-07-10
Add ovirt provider e2e test

bennyz:
A very raw attempt to create an e2e test that adds an oVirt provider and checks that it is added successfully. The relies on existing environments, and expects the following environment variables to be set:
```
OVIRT_USERNAME
OVIRT_PASSWORD
OVIRT_CACERT
OVIRT_URL
```

The test can be run using the new target in the Makefile:
```shell
$ make e2e
```


ahadas:
@liranr23 works on that now, will get in to kubev2v/forklift

PR #439, opened 2022-07-11
Add correlation id to oVirt provider requests

bennyz:
Setting the correlation id would allow us to see which migrationtriggered an action on oVirt and make it easier to debug.

Example:
```
2022-07-11 17:28:16,138+03 INFO  [org.ovirt.engine.core.bll.snapshots.CreateSnapshotForVmCommand] (default task-116) [copy-of-plan-1657549684863] Lock Acquired to object 'EngineLock:{exclusiveLocks='[27bc786a-f52b-4f3b-b3da-25a4dced467d=VM]', sharedLocks=''}'
```


ahadas:
I didn't see if this covers all places but it helped us quite a few times with rhv bugs, so sounds like a great idea!

PR #442, opened 2022-07-18
hack: add deploy script

bennyz:
Add a script to build the image and deploy it to the local openshift registry (or any other registry).

By default uses the forklift-konveyor namespace and forklift-controller:latest image name. The upload will be done to the default openshift registry and the script will edit the CSV to use the image from registry instead of the image from quay.

### DIFF
--- a/hack/deploy.sh
+++ b/hack/deploy.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+set -e
+
+function help() {
+   echo "Deploy a custom image of the forklift-controller to a custom registry."
+   echo
+   echo "Syntax: deploy.sh [-h|n|i|r]"
+   echo "options:"
+   echo "h     Print this help"
+   echo "n     Namespace to use (Default: konveyor-forklift)"
+   echo "i     Image name       (Default: forklift-controller:latest)"
+   echo "r     Registry         (Default: uses default openshift route)"
+   echo
+}
+
+while getopts n:i:r:h flag
+do
+    case "${flag}" in
+	n) namespace=${OPTARG};;
+        i) image=${OPTARG};;
+	r) registry=${OPTARG};;
+	h) help
+	   exit;;
+    esac
+done
+
+NAMESPACE=${namespace:-konveyor-forklift}
+IMG=${image:-forklift-controller:latest}
+REGISTRY=${registry:-"$(oc get route/default-route -n openshift-image-registry -o=jsonpath='{.spec.host}')/openshift"}
+
+podman login --tls-verify=false -u unused -p $(oc whoami -t) ${REGISTRY}
+podman build -t ${REGISTRY}/${IMG} .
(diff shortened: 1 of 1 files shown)


hack/deploy.sh
```diff
@@ -0,0 +1,38 @@
...
+
+NAMESPACE=${namespace:-konveyor-forklift}
+IMG=${image:-forklift-controller:latest}
+REGISTRY=${registry:-"$(oc get route/default-route -n openshift-image-registry -o=jsonpath='{.spec.host}')/openshift"}
+
+podman login --tls-verify=false -u unused -p $(oc whoami -t) ${REGISTRY}
+podman build -t ${REGISTRY}/${IMG} .
+podman push --tls-verify=false ${REGISTRY}/${IMG}
+
+CSV=$(oc get csv -n konveyor-forklift -o custom-columns=":metadata.name" --no-headers=true)
```
liranr23, line 35: seems it requires oc to be logged in, i think it worth a mention in the `help()` (requires oc) for login you may add option to connect oc here as well. or, just add it as a requirement in the `help()`.

ahadas:
@bennyz how about taking [the script from openshift-logging](https://github.com/openshift/cluster-logging-operator/blob/master/hack/deploy-image.sh) as a reference? they incorporated it in the makefile and it worked pretty well few years ago :)

bennyz:
> @bennyz how about taking [the script from openshift-logging](https://github.com/openshift/cluster-logging-operator/blob/master/hack/deploy-image.sh?rgh-link-date=2022-07-20T10%3A21%3A28Z) as a reference? they incorporated it in the makefile and it really served me well few years ago :) (and while searching for it, I noticed they replaced the plugin for fluentd like I proposed back then, nice!)

At first this script looked completely alien to me :) but I think I understand why it would not be straightforward to do the same here, the openshift logging image is already in the internal registry, so pushing using the same tag just works out of the box. In our case the image is on quay.io, so we have to do some extra work to manipulate the image used by controller deployment...
I'm not sure about the port forwarding though

ahadas:
@bennyz sure, some parts would be different but things like setting the port-forwarding which I think is needed here as well in order to be able to access the registry, setting the image tag per branch, executing the script using the makefile (no need for the help section, in my opinion), separating out the part that builds the image from this script - things like that

bennyz:
> @bennyz right, they are similar already which is a good thing and it makes sense that some parts would be different but things like setting the port-forwarding which I think is needed here as well in order to be able to access the registry, setting the image tag per branch, executing the script using the makefile (no need for the help section, in my opinion), separating out the part that builds the image from this script - I meant such things

yeah, I guess we can separate building out and make it more focused. Regarding the port forwarding, the reason I don't understand why it's needed is that it works without it for me 😅

ahadas:
I believe this is not needed anymore with the changes we've done in kubev2v/forklift

PR #443, opened 2022-07-18
Bug 2055201: Avoid starting importer before snapshots are ready to transfer.

mrnold:
Bug [2055201](https://bugzilla.redhat.com/show_bug.cgi?id=2055201) shows frequent importer pod restarts for warm migrations from oVirt, and one of the reasons given by the oVirt API is "Cannot transfer Virtual Disk. Snapshot is currently being created for VM".

It should be easy enough to clean up these restarts by having the forklift controller wait for the snapshot to be marked ready, but I'm not sure if I have the right approach in this pull request? I have marked it a draft in case adding to the state machine isn't really the right thing to do, and because the VMware part isn't really filled out (although VMware doesn't seem to have the same problem).

### DIFF
--- a/pkg/controller/plan/adapter/ovirt/client.go
+++ b/pkg/controller/plan/adapter/ovirt/client.go
@@ -75,6 +75,26 @@ func (r *Client) RemoveSnapshots(vmRef ref.Ref, precopies []planapi.Precopy) (er
 }
 
 //
+// Check if a snapshot is ready to transfer, to avoid importer restarts
+func (r *Client) CheckSnapshotReady(vmRef ref.Ref, snapshot string) (ready bool, err error) {
+	_, vmService, err := r.getVM(vmRef)
+	if err != nil {
+		return
+	}
+	snapsService := vmService.SnapshotsService()
+	snapService := snapsService.SnapshotService(snapshot)
+	snap, err := snapService.Get().Query("correlation_id", r.Migration.Name).Send()
+	if err != nil {
+		err = liberr.Wrap(err)
+	}
+	status := snap.MustSnapshot().MustSnapshotStatus()
+	if status == ovirtsdk.SNAPSHOTSTATUS_OK {
+		ready = true
+	}
+	return
+}
+
+//
 // Set DataVolume checkpoints.
 func (r *Client) SetCheckpoints(vmRef ref.Ref, precopies []planapi.Precopy, datavolumes []cdi.DataVolume, final bool) (err error) {
 	n := len(precopies)
--- a/pkg/controller/plan/adapter/ovirt/client.go
+++ b/pkg/controller/plan/adapter/ovirt/client.go
@@ -95,22 +95,19 @@ func (r *Client) CheckSnapshotReady(vmRef ref.Ref, snapshot string) (ready bool,
 		err = liberr.Wrap(err)
 		return
 	}
-	if len(jobs) > 1 {
-		err = liberr.New("Multiple jobs found for correlation ID %s", correlationID)
-		return
-	} else if len(jobs) < 1 {
+	if len(jobs) < 1 {
 		err = liberr.New("No jobs found for correlation ID %", correlationID)
 		return
 	}
-	job := jobs[0]
-	switch job.MustStatus() {
-	case ovirtsdk.JOBSTATUS_FAILED, ovirtsdk.JOBSTATUS_ABORTED:
-		err = liberr.New("Snapshot creation failed! Correlation ID is %s", correlationID)
-		ready = false
-	case ovirtsdk.JOBSTATUS_STARTED, ovirtsdk.JOBSTATUS_UNKNOWN:
-		ready = false
-	case ovirtsdk.JOBSTATUS_FINISHED:
-		ready = true
+	ready = true
+	for _, job := range jobs {
+		switch job.MustStatus() {
+		case ovirtsdk.JOBSTATUS_FAILED, ovirtsdk.JOBSTATUS_ABORTED:
+			err = liberr.New("Snapshot creation failed! Correlation ID is %s", correlationID)
+			ready = false
+		case ovirtsdk.JOBSTATUS_STARTED, ovirtsdk.JOBSTATUS_UNKNOWN:
+			ready = false
+		}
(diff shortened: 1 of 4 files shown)


ahadas:
This change would definitely improve the current state but it won't prevent the problem from happening - the reason for getting "Cannot transfer Virtual Disk. Snapshot is currently being created for VM" is that the VM is locked and the VM is unlocked after changing the status of the snapshot. It would be better if we can check the state of the create-snapshot command using its correlation id
CC @bennyz

mrnold:
Okay, I think I see how to do that. Should I create a unique correlation ID for each snapshot? Or just keep using the migration name as the correlation ID, and make sure all the existing correlated jobs are finished?

ahadas:
> Okay, I think I see how to do that. Should I create a unique correlation ID for each snapshot? Or just keep using the migration name as the correlation ID, and make sure all the existing correlated jobs are finished?

In #439 the migration name is as the correlation ID but now I wonder if that is the right thing to do - that correlation-ID is supposed to be unique per-job, I'm not sure what would happen if one create-snapshot operation with that correlation-ID fails and the following one with that correlation-ID succeeds, what would we get when querying the status of the job associated with that correlation ID? I think it may be an arbitrary result. @bennyz did you consider such flow?

bennyz:
> > Okay, I think I see how to do that. Should I create a unique correlation ID for each snapshot? Or just keep using the migration name as the correlation ID, and make sure all the existing correlated jobs are finished?
> 
> In #439 the migration name is as the correlation ID but now I wonder if that is the right thing to do - that correlation-ID is supposed to be unique per-job, I'm not sure what would happen if one create-snapshot operation with that correlation-ID fails and the following one with that same correlation-ID succeeds, what would we get when querying the status of the job associated with that correlation ID? I think it may be an arbitrary result. @bennyz did you consider such flow?

well I kind of did and didn't, since correlation id was not used for anything, I just added for tracing purposes in the engine log and not to check  job completion. But if we're going to use it to wait for operations, yes, it would need to be adjusted

mrnold:
I ended up using a unique correlation ID for each snapshot, based on the plan name, the VM UUID, and the index of the snapshot in the list of precopies. I had some trouble with correlation IDs longer than a UUID, so I hashed it with MD5. It's kind of awkward but seems to be reliably getting the right job status.

ahadas (approved):
lgtm
unrelated to your change - I hope we'll be able to align the warm migration from vsphere and warm migration from rhv at some point as cbt was also added to rhv 4.4

mrnold:
> lgtm unrelated to your change - I hope we'll be able to align the warm migration from vsphere and warm migration from rhv at some point as cbt was also added to rhv 4.4

Yes, this would be a big improvement.

mrnold:
This should be ready to go now.

pkg/controller/plan/migration.go
```diff
@@ -706,11 +709,34 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
 
 		switch vm.Phase {
```
mansam, line 710: If you add the `WaitFor...` phases to the warm migration itinerary on line 103, you can replace this case with `vm.Phase = r.next(vm.Phase)`.

For example:

```
{Name: CreateInitialSnapshot},
{Name: WaitForInitialSnapshot},
{Name: CreateDataVolumes},
```

pkg/controller/plan/migration.go
```diff
@@ -706,11 +709,34 @@ func (r *Migration) execute(vm *plan.VMStatus) (err error) {
...
+		}
+		snapshot := vm.Warm.Precopies[len(vm.Warm.Precopies)-1].Snapshot
+		ready, err := r.provider.CheckSnapshotReady(vm.Ref, snapshot)
+		if err != nil {
+			step.AddError(err.Error())
+			err = nil
+			break
+		}
+		if ready {
+			switch vm.Phase {
```
mansam, line 732: I think this can be replaced with `vm.Phase = r.next(vm.Phase)` as well if the `WaitFor...` phases are added to the right spots in the warm itinerary.

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -73,6 +81,39 @@ func (r *Client) RemoveSnapshots(vmRef ref.Ref, precopies []planapi.Precopy) (er
...
+		err = liberr.Wrap(err)
+		return
+	}
+	jobs, err := r.getJobs(correlationID)
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+	if len(jobs) > 1 {
+		err = liberr.New("Multiple jobs found for correlation ID %s", correlationID)
```
bennyz, line 98: I wouldn't make this an error, for snapshots there is one job, but it is not unlikely to have multiple jobs for the same correlation id

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -73,6 +81,39 @@ func (r *Client) RemoveSnapshots(vmRef ref.Ref, precopies []planapi.Precopy) (er
...
+		return
+	} else if len(jobs) < 1 {
+		err = liberr.New("No jobs found for correlation ID %", correlationID)
+		return
+	}
+	job := jobs[0]
```
bennyz, line 104: I'd check all return jobs, this is what we usually do when we use correlation id to check for completion

mansam (changes requested):
Looks good, just needs an adjustment to the itinerary.

pkg/controller/plan/adapter/ovirt/client.go
mrnold: Okay, I removed this check.

pkg/controller/plan/adapter/ovirt/client.go
mrnold: I changed this around a bit to consider the status of all the listed jobs.

pkg/controller/plan/migration.go
mrnold: Okay, I think I have flattened this one and the next one the way you wanted.

mansam (approved):
LGTM

PR #444, opened 2022-07-18
Consolidate provider related things

ahadas:
This PR attempts to take the provider-type's related things that scattered in the code and move them close to where the provider types are defined to ease adding a new provider


ahadas:
@mansam changed the location of RequiresConversion as you proposed, can you please take another look?

mansam (approved):
LGTM

liranr23 (approved):

pkg/apis/forklift/v1beta1/provider.go
```diff
@@ -27,17 +27,30 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
+type ProviderType string
```
jortel, line 30: Please do NOT create types from built-in types.  This adds no value and just introduces cruft such as requiring .String() everywhere.  We deliberately avoided this in our coding conventions.
Please revert all but the new method.

PR #445, opened 2022-07-22
host validation: change vmware fail error msg

mnecas:
Bz: https://bugzilla.redhat.com/show_bug.cgi?id=1958826
Issue: 
When selecting the migration network on the VMware host and entering the wrong credentials the throwen error is unclear.

### DIFF
--- a/pkg/controller/host/validation.go
+++ b/pkg/controller/host/validation.go
@@ -86,7 +86,7 @@ func (r *Reconciler) validate(host *api.Host) error {
 			Status:   True,
 			Reason:   Completed,
 			Category: Advisory,
-			Message:  "The host has been validated.",
+			Message:  "The host inventory reference, IP address, and credentials have been validated.",
 		})
 	err = r.testConnection(host)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/host/validation.go
```diff
@@ -86,7 +86,7 @@ func (r *Reconciler) validate(host *api.Host) error {
 			Status:   True,
 			Reason:   Completed,
 			Category: Advisory,
-			Message:  "The host has been validated.",
+			Message:  "The host resources has been validated.",
```
rokkbert, line 89: "resources" is plural, so maybe it should be "resources have been validated"?

pkg/controller/host/validation.go
```diff
@@ -86,7 +86,7 @@ func (r *Reconciler) validate(host *api.Host) error {
 			Status:   True,
 			Reason:   Completed,
 			Category: Advisory,
-			Message:  "The host has been validated.",
+			Message:  "The host resources have been validated.",
```
ahadas, line 89: it was suggested in the bug to elaborate on what's being validated (Host inventory reference, IP address, and credentials) - I'm not sure the term 'host resources' would be clear to users

michalskrivanek (approved):

PR #446, opened 2022-07-24
Recheck provider periodically

bennyz:
Currently once a provider has been reconciled it will no longer be checked, unless changes have been made to it. This may lead to a situation where a provider that is no longer active will appear as ready.

This patch reconciles every 10 minutes to make sure it is still functional, but should probably be made configurable if this approach is acceptable

### DIFF
--- a/pkg/controller/provider/controller.go
+++ b/pkg/controller/provider/controller.go
@@ -18,6 +18,11 @@ package provider
 
 import (
 	"context"
+	"os"
+	"path/filepath"
+	"sync"
+	"time"
+
 	libcnd "github.com/konveyor/controller/pkg/condition"
 	liberr "github.com/konveyor/controller/pkg/error"
 	libfb "github.com/konveyor/controller/pkg/filebacked"
@@ -36,15 +41,12 @@ import (
 	core "k8s.io/api/core/v1"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apiserver/pkg/storage/names"
-	"os"
-	"path/filepath"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
-	"sync"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


ahadas (commented):
besides things like defining this 10min constant like SlowReQ,
1. are we prepared for such a state transition from 'ready' to non-ready state elsewhere? IIRC, I couldn't define a plan until the provider became 'ready' but if I have a plan and then the provider is set as non-ready and without condition ConnectionTestSucceeded - would we block the execution of the plan?
2. should we clear the inventory so the user won't see stale data? (and yes, I guess the current state is not better than with this change)
3. is it possible to change the ConnectionTestSucceeded condition and trigger the reconcile procedure when we fail to communicate with the provider as part of refreshing the inventory? what's the frequency of reaching out to the provider in order to update the inventory?

bennyz:
> besides things like defining this 10min constant like SlowReQ,

I actually thought about putting this in the CRD so it's configurable by the user

> 1. are we prepared for such a state transition from 'ready' to non-ready state elsewhere? IIRC, I couldn't define a plan until the provider became 'ready' but if I have a plan and then the provider is set as non-ready and without condition ConnectionTestSucceeded - would we block the execution of the plan?

Yes, this is another question, this PR is for the case where the provider just "sits there", it probably makes sense to force reconciliation when dependent resources' controllers run (plan, migration), if it's not done already 

> 2. should we clear the inventory so the user won't see stale data? (and yes, I guess the current state is not better than with this change)

I believe we should, I'll check how it goes, might be better suited for a follow-up PR

> 3. is it possible to change the ConnectionTestSucceeded condition and trigger the reconcile procedure when we fail to communicate with the provider as part of refreshing the inventory? what's the frequency of reaching out to the provider in order to update the inventory?

This was actually my initial thought (and the original PR was much larger than this), but ultimately it seemed more complicated because and a little out of place because the container does not interact with the k8s api, but I'm open to any suggestions :)
The interval is 10 seconds:
https://github.com/konveyor/forklift-controller/blob/a4773012cbdfcfeafb394de84b8c6fdd703deb46/pkg/controller/provider/container/ovirt/collector.go#L28

michalskrivanek:
I think we do need a visual feedback and perhaps control as well.
10 minutes is quite slow, but OTOH we don't want to poke API too frequently forever.
As a separate PR  of course

jortel (changes requested):
This should NOT be done with a re-queue because it will cause all other resources with references to the provider to reconcile (including the plan which will also cascade).  This needs to be done with a custom _WatchSource_ that checks the connection and queues a reconcile when the connection health changes.

See: https://github.com/konveyor/mig-controller/blob/master/pkg/controller/migcluster/remotecluster.go#L19

ahadas:
@bennyz let's continue this in kubev2v/forklift

PR #447, opened 2022-07-27
Add ConnectionAuthFailed for one try auth check 

mnecas:
Issue: when a user enters the correct username for the provider but incorrect password it will try to connect multiple times to the engine. Which locks the user account.
Fix: Adding a special case of condition `ConnectionAuthFailed` and finish the reconciliation.
Requirement: https://github.com/konveyor/controller/pull/104

### DIFF
--- a/pkg/controller/provider/container/ovirt/client.go
+++ b/pkg/controller/provider/container/ovirt/client.go
@@ -144,13 +144,13 @@ func (r *Client) auth() (user string) {
 
 //
 // Get system.
-func (r *Client) system() (s *System, err error) {
+func (r *Client) system() (s *System, status int, err error) {
 	err = r.connect()
 	if err != nil {
 		return
 	}
 	system := &System{}
-	status, err := r.client.Get(r.url, system)
+	status, err = r.client.Get(r.url, system)
 	if err != nil {
 		return
 	}
--- a/pkg/controller/provider/container/vsphere/collector.go
+++ b/pkg/controller/provider/container/vsphere/collector.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	liburl "net/url"
 	"path"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -497,7 +498,7 @@ func (r *Collector) watch() (list []*libmodel.Watch) {
 
 //
 // Build the client.
-func (r *Collector) connect(ctx context.Context) (int, error) {
+func (r *Collector) connect(ctx context.Context) (status int, err error) {
 	r.close()
 	url, err := liburl.Parse(r.url)
 	if err != nil {
@@ -518,7 +519,10 @@ func (r *Collector) connect(ctx context.Context) (int, error) {
 	}
 	err = r.client.Login(ctx, url.User)
 	if err != nil {
-		return http.StatusUnauthorized, liberr.Wrap(err)
+		if strings.Contains(err.Error(), "Cannot complete login due to an incorrect user name or password") {
+			return http.StatusUnauthorized, liberr.Wrap(err)
+		}
+		return 0, liberr.Wrap(err)
 	}
 
 	return http.StatusOK, nil
--- a/pkg/controller/provider/container/vsphere/collector.go
+++ b/pkg/controller/provider/container/vsphere/collector.go
@@ -2,6 +2,11 @@ package vsphere
 
 import (
 	"context"
+	"net/http"
+	liburl "net/url"
+	"path"
+	"time"
+
 	"github.com/go-logr/logr"
 	liberr "github.com/konveyor/controller/pkg/error"
 	libmodel "github.com/konveyor/controller/pkg/inventory/model"
@@ -17,9 +22,6 @@ import (
 	"github.com/vmware/govmomi/vim25/types"
 	core "k8s.io/api/core/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
-	liburl "net/url"
-	"path"
-	"time"
 )
 
 //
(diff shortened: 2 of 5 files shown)


pkg/controller/provider/controller.go
```diff
@@ -231,6 +231,9 @@ func (r Reconciler) Reconcile(request reconcile.Request) (result reconcile.Resul
 			"Waiting connection tested or inventory created.")
 		result.RequeueAfter = base.SlowReQ
 	}
+	if provider.Status.HasCondition(ConnectionAuthFailed) {
+		result.RequeueAfter = 0
```
ahadas, line 235: does it change when we post an update to the provider?

ahadas (commented):
I think that with a proper admission webhook this would handle just the corner case of changing the credentials after defining the provider..

pkg/controller/provider/controller.go
mnecas: yes, when we change some things on which we have watch, the Reconcile gets triggered again.
https://github.com/konveyor/forklift-controller/blob/76ee1608b4850a0d47e92b508627b4e94847e29c/pkg/controller/provider/controller.go#L101-L114

ahadas: ok, makes sense

pkg/controller/provider/controller.go
```diff
@@ -232,6 +232,13 @@ func (r Reconciler) Reconcile(request reconcile.Request) (result reconcile.Resul
 		result.RequeueAfter = base.SlowReQ
 	}
 
+	// Stop reconciliation when auth fails
+	if provider.Status.HasCondition(ConnectionAuthFailed) {
+		result.RequeueAfter = 0
+		err = nil
+		return
```
mnecas, line 239: With PR https://github.com/konveyor/forklift-controller/pull/453 I can move this check right after the validation.
 https://github.com/konveyor/forklift-controller/blob/367d2e9e2378ba7419881aff80fd32472a4b8538/pkg/controller/provider/controller.go#L186-L189

pkg/controller/provider/container/ovirt/client.go
```diff
@@ -144,13 +144,13 @@ func (r *Client) auth() (user string) {
 
 //
 // Get system.
-func (r *Client) system() (s *System, err error) {
+func (r *Client) system() (s *System, status int, err error) {
```
rokkbert, line 147: It seems that the return value _s_ should be called _system_. I don't think it is really used anywhere, but maybe it could just be fixed at the same time.

pkg/controller/provider/container/vsphere/collector.go
```diff
@@ -495,11 +497,11 @@ func (r *Collector) watch() (list []*libmodel.Watch) {
 
 //
 // Build the client.
-func (r *Collector) connect(ctx context.Context) error {
+func (r *Collector) connect(ctx context.Context) (int, error) {
```
rokkbert, line 500: If the new int had a name it would be clear what it is even without reading the body of the function. Maybe httpStatus or something like that?

pkg/controller/provider/container/vsphere/collector.go
```diff
@@ -508,18 +510,18 @@ func (r *Collector) connect(ctx context.Context) error {
...
+		return 0, liberr.Wrap(err)
 	}
 	r.client = &govmomi.Client{
 		SessionManager: session.NewManager(vimClient),
 		Client:         vimClient,
 	}
 	err = r.client.Login(ctx, url.User)
 	if err != nil {
-		return liberr.Wrap(err)
+		return http.StatusUnauthorized, liberr.Wrap(err)
```
rokkbert, line 521: Is it certain that this is the only error we could get from Login? If the Login fails for other reasons (network?) then it might be misleading to turn that into a StatusUnauthorized.

pkg/controller/provider/container/vsphere/collector.go
```diff
@@ -508,18 +511,21 @@ func (r *Collector) connect(ctx context.Context) error {
...
+		return
 	}
 	r.client = &govmomi.Client{
 		SessionManager: session.NewManager(vimClient),
 		Client:         vimClient,
 	}
 	err = r.client.Login(ctx, url.User)
 	if err != nil {
-		return liberr.Wrap(err)
+		if strings.Contains(err.Error(), "Cannot complete login due to an incorrect user name or password") {
```
mnecas, line 522: Really don't like this solution but the only alternative I can think of is to rewrite the communication for VMware to basic http requests. Which would need to include the requests for the ca token to validate the thumbprint. Not sure if it would be worth this issue.
(I read docs for the govmomi and could not find any function which could return the http status code of url.)

michalskrivanek: lgtm

pkg/controller/provider/validation.go
```diff
@@ -268,6 +270,19 @@ func (r *Reconciler) testConnection(provider *api.Provider, secret *core.Secret)
...
 	} else {
+		if status == http.StatusUnauthorized {
+			provider.Status.SetCondition(
+				libcnd.Condition{
+					Type:     ConnectionAuthFailed,
+					Status:   True,
+					Reason:   Tested,
+					Category: Critical,
+					Message: fmt.Sprintf(
+						"Connection auth failed, failed: %s",
```
michalskrivanek, line 281: "... failed, failed: " ?

michalskrivanek (commented):
please also add a code comment regarding the special behavior for 401

pkg/controller/provider/validation.go
```diff
@@ -257,6 +257,10 @@ func (r *Reconciler) testConnection(provider *api.Provider, secret *core.Secret)
 		return nil
 	}
 	rl := container.Build(nil, provider, secret)
+	// Status code returns
+	// 0 - ignoring (unrealted connection fail)
+	// 200 - ok
+	// 401 - unauthorized
```
michalskrivanek, line 263: I didn't mean to document http codes but rather to explain what are we doing and why do we want this special behavior for "Unauthorized"

mnecas:
Patch was migrated to https://github.com/kubev2v/forklift/pull/31

PR #448, opened 2022-08-01
Change VM names comply with RFC 1123 DNS Subdomain

bkhizgiy:
This PR changes VMs names to match the DNS1123 RFC conventions,
As a first step, the names will be automatically adjusted to make the VMs ready for migration without user actions.
A newly generated name will be created and validated with the destination, if exists, will add the first 4 chars from the ID to the new name to avoid dups. 
A new annotation with the previous name and ID will be added to the imported VM.

Currently, implementation checks the following:
- contain at most 63 characters
- contain only lowercase alphanumeric characters or '-'
- start with an alphanumeric character
- end with an alphanumeric character

In case the VM name is completely deleted after the adjustments a new name in the format `vm-$ID` will be given.

https://bugzilla.redhat.com/show_bug.cgi?id=1959883

This [fix](https://github.com/konveyor/forklift-validation/pull/47) changes the validation displayed message.

### DIFF
--- a/pkg/controller/plan/kubevirt.go
+++ b/pkg/controller/plan/kubevirt.go
@@ -4,6 +4,14 @@ import (
 	"context"
 	"encoding/xml"
 	"fmt"
+	"math/rand"
+	"path"
+	"regexp"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+
 	template "github.com/openshift/api/template/v1"
 	"github.com/openshift/library-go/pkg/template/generator"
 	"github.com/openshift/library-go/pkg/template/templateprocessing"
@@ -12,14 +20,9 @@ import (
 	"k8s.io/apimachinery/pkg/conversion"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	k8svalidation "k8s.io/apimachinery/pkg/util/validation"
 	cnv "kubevirt.io/client-go/api/v1"
 	libvirtxml "libvirt.org/libvirt-go-xml"
-	"math/rand"
-	"path"
-	"sort"
-	"strconv"
-	"strings"
-	"time"
 
 	libcnd "github.com/konveyor/controller/pkg/condition"
 	liberr "github.com/konveyor/controller/pkg/error"
@@ -30,6 +33,7 @@ import (
 	core "k8s.io/api/core/v1"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	cdi "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -44,6 +48,10 @@ const (
 	// Contains validations for a Kubevirt VM. Needs to be removed when
 	// creating a VM from a template.
 	AnnKubevirtValidations = "vm.kubevirt.io/validations"
+	//  Original VM name on source (value=vmOriginalName)
+	AnnOriginalName = "original-name"
+	//  Original VM name on source (value=vmOriginalID)
+	AnnOriginalID = "original-ID"
 )
 
 // Labels
@@ -696,6 +704,27 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
 		return
 	}
 
+	//If the VM name is not valid according to DNS1123 labeling
+	//convention it will be automatically changed.
+	var originalName string
+
+	if errs := k8svalidation.IsDNS1123Label(vm.Name); len(errs) > 0 {
+		originalName = vm.Name
+
+		generatedName := changeVmName(vm.Name, vm.ID)
+		nameExist, errName := r.checkIfVmNameExist(generatedName)
+		if errName != nil {
+			err = liberr.Wrap(errName)
+			return
+		}
+		if nameExist {
+			generatedName = generatedName + "-" + vm.ID[:4]
+		}
+		vm.Name = generatedName
+		r.Log.Info("VM name was not compatible with DNS1123 RFC, changing to ", "vm",
+			vm.Name)
+	}
+
 	var ok bool
 	object, ok = r.vmTemplate(vm)
 	if !ok {
@@ -704,6 +733,13 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
 			vm.String())
 		object = r.emptyVm(vm)
 	}
+	//Add the original name and ID info to the VM annotations
+	if len(originalName) > 0 {
+		annotations := make(map[string]string)
+		annotations[AnnOriginalName] = originalName
+		annotations[AnnOriginalID] = vm.ID
+		object.ObjectMeta.Annotations = annotations
+	}
 	running := false
 	object.Spec.Running = &running
 
@@ -711,7 +747,6 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
 	if err != nil {
 		return
 	}
-
 	return
 }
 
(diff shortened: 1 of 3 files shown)


pkg/controller/plan/kubevirt.go
```diff
@@ -239,6 +242,11 @@ func (r *KubeVirt) EnsureVM(vm *plan.VMStatus) (err error) {
 		err = liberr.Wrap(err)
 		return
 	}
+	//If the VM name is not valid according to DNS1123 labeling
+	//convention it will be automatically changed.
+	if len(k8svalidation.IsDNS1123Label(vm.Name)) > 0 {
+		newVM.Name = changeVmNameDNS1123(vm.Name, vm.ID)
+	}
```
liranr23, line 249: i think this part should be within the `virtualMachine()`

pkg/controller/plan/kubevirt.go
```diff
@@ -1389,3 +1397,28 @@ func vmOwnerReference(vm *cnv.VirtualMachine) (ref meta.OwnerReference) {
 	}
 	return
 }
+
+//changes VM name to match DNS1123 RFC convention.
+func changeVmNameDNS1123(currName string, vmID string) string {
+
+	var DNS1123LabelMaxLength int = 63
```
liranr23, line 1404: i think it will be nicer to remove the `DNS1123` from variable and even from the function. The function description tells what the convention we are based on.

pkg/controller/plan/validation.go
```diff
@@ -317,8 +318,8 @@ func (r *Reconciler) validateVM(plan *api.Plan) error {
 		Type:     NameNotValid,
 		Status:   True,
 		Reason:   NotValid,
-		Category: Critical,
-		Message:  "Target VM name not valid.",
+		Category: Warn,
+		Message:  "Target VM name not valid, will be automatically changed.",
```
liranr23, line 322: maybe here we can say based on DNS1123 RFC convention

pkg/controller/plan/kubevirt.go
```diff
@@ -239,6 +242,11 @@ func (r *KubeVirt) EnsureVM(vm *plan.VMStatus) (err error) {
 		err = liberr.Wrap(err)
 		return
 	}
+	//If the VM name is not valid according to DNS1123 labeling
+	//convention it will be automatically changed.
+	if len(k8svalidation.IsDNS1123Label(vm.Name)) > 0 {
```
ahadas, line 247: how about `if errs := IsDNS1123Label(label); len(errs) > 0 {`, I find it more readable than checking the length of a return value of a function that starts with 'Is'

pkg/controller/plan/kubevirt.go
```diff
@@ -239,6 +242,11 @@ func (r *KubeVirt) EnsureVM(vm *plan.VMStatus) (err error) {
...
+		newVM.Name = changeVmNameDNS1123(vm.Name, vm.ID)
```
ahadas, line 248: wouldn't it be useful to log the name change here?

pkg/controller/plan/kubevirt.go
ahadas: yeah, makes sense to either move it there or change the vm before calling `virtualMachine(...)`

pkg/controller/plan/validation.go
ahadas: or: Target VM name does not comply with DNS1123 RFC, ...

pkg/controller/plan/kubevirt.go
bkhizgiy: Agreed, changed according to your suggestion

pkg/controller/plan/kubevirt.go
bkhizgiy: agreed, moved to `virtualMachine()`

liranr23 (approved):

pkg/controller/plan/kubevirt.go
```diff
@@ -696,13 +701,28 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
...
 	object, ok = r.vmTemplate(vm)
 	if !ok {
 		r.Log.Info("Building VirtualMachine without template.",
 			"vm",
 			vm.String())
-		object = r.emptyVm(vm)
+		if len(originalName) == 0 {
+			object = r.emptyVm(vm)
+		} else {
+			object = r.emptyVmwithOriginalNameLable(vm, originalName)
```
liranr23, line 724: typo `Lable` -> `Label`.

pkg/controller/plan/kubevirt.go
```diff
@@ -786,6 +806,23 @@ func (r *KubeVirt) emptyVm(vm *plan.VMStatus) (virtualMachine *cnv.VirtualMachin
...
+// Create empty VM definition with origin name lable.
+func (r *KubeVirt) emptyVmwithOriginalNameLable(vm *plan.VMStatus, originalName string) (virtualMachine *cnv.VirtualMachine) {
+	virtualMachine = &cnv.VirtualMachine{
+		TypeMeta: meta.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "VirtualMachine",
+		},
+		ObjectMeta: meta.ObjectMeta{
+			Namespace: r.Plan.Spec.TargetNamespace,
+			Labels:    r.vmOriginalNameLables(originalName, vm.Ref),
```
liranr23, line 818: typo `Lables` -> `Labels`.

pkg/controller/plan/kubevirt.go
```diff
@@ -786,6 +806,23 @@ func (r *KubeVirt) emptyVm(vm *plan.VMStatus) (virtualMachine *cnv.VirtualMachin
 	return
 }
 
+// Create empty VM definition with origin name lable.
```
liranr23, line 809: `lable` typo

pkg/controller/plan/kubevirt.go
```diff
@@ -786,6 +806,23 @@ func (r *KubeVirt) emptyVm(vm *plan.VMStatus) (virtualMachine *cnv.VirtualMachin
...
+func (r *KubeVirt) emptyVmwithOriginalNameLable(vm *plan.VMStatus, originalName string) (virtualMachine *cnv.VirtualMachine) {
```
liranr23, line 810: how about having it in a single place (e.g `emptyVm` above), getting also `originalName`. you can create the vm with `ObjectMeta` that doesn't contain the `Labels` and add it at the end based on the `if`. Another option would be to have 2 functions creating the `ObjectMeta`.

pkg/controller/plan/kubevirt.go
```diff
@@ -54,6 +57,8 @@ const (
 	kPlan = "plan"
 	// VM label (value=vmID)
 	kVM = "vmID"
+	//  Original VM name on source (value=vmOriginalName)
+	kOriginalName = "original-name"
```
liranr23, line 61: originalName?

pkg/controller/plan/kubevirt.go
bkhizgiy: good idea, changed

pkg/controller/plan/kubevirt.go
```diff
@@ -704,6 +738,13 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
 			vm.String())
 		object = r.emptyVm(vm)
 	}
+	//Add the original name and ID info to the VM annotations
+	if len(originalName) > 0 {
+		annotations := make(map[string]string)
+		annotations[AnnOriginalName] = originalName
+		annotations[AnnOriginalID] = vm.ID
+		object.ObjectMeta.Annotations = annotations
```
liranr23, line 746: wouldn't it override the annotations to have only AnnOriginalName and AnnOriginalID?

shouldn't it be something like:
```
annotations := object.ObjectMeta.Annotations
annotations[AnnOriginalName] = originalName
annotations[AnnOriginalID] = vm.ID
object.ObjectMeta.Annotations = annotations
```
(maybe nil check required for the first line, if it's nil then to make the map).

pkg/controller/plan/kubevirt.go
```diff
@@ -696,6 +707,29 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
...
 
+	//If the VM name is not valid according to DNS1123 labeling
+	//convention it will be automatically changed.
+	var originalName string
+
+	if errs := k8svalidation.IsDNS1123Label(vm.Name); len(errs) > 0 {
+		originalName = vm.Name
+
+		generatedName := changeVmName(vm.Name, vm.ID)
+		r.mutex.Lock()
```
liranr23, line 718: why the mutex is required? maybe a comment on that?

pkg/controller/plan/kubevirt.go
bkhizgiy: At this point the annotation part does not exist, you can see in `emptyVm()` that it's not created yet, so here I am just initializing them, and later on in the flow, the rest of the annotations are added so there's no overriding (also been tested and all the annotation remains).

pkg/controller/plan/kubevirt.go
bkhizgiy: sure will add a comment, basically this function checks if the new generated VM name exists on the destination and if so adds the ID to the name, I've added a mutex for the case that two or more VMS executing this part at parallel and the Vm list is not updated yet and after renaming both of the names will be the same (`Vm--` and `--vm` for example both will change to `vm`) but the function my miss that point since the full VM list is not updated yet, so to avoid any issues I've added a mutex so only one VM at a time can retrieve the data.

bkhizgiy: mutex not needed, the issue was different and solved by comparing the name with the VMS from the same plan.

pkg/controller/plan/kubevirt.go
```diff
@@ -704,14 +733,20 @@ func (r *KubeVirt) virtualMachine(vm *plan.VMStatus) (object *cnv.VirtualMachine
 			vm.String())
 		object = r.emptyVm(vm)
 	}
+	//Add the original name and ID info to the VM annotations
+	if len(originalName) > 0 {
+		annotations := make(map[string]string)
+		annotations[AnnOriginalName] = originalName
+		annotations[AnnOriginalID] = vm.ID
```
ahadas, line 740: I'd always set the original-id, it can be handy for debugging also when we don't rename the vm

ahadas (commented):
lgtm, the comment inside is on something that is not specific to this fix

ahadas:
@bkhizgiy how about posting this in https://github.com/kubev2v/forklift? we don't plan to backport that..

PR #452, opened 2022-08-03
VM mapping: map IDE to Sata

mnecas:
the user may not have virtio drivers for windows guests, it's better to convert them to SATA instead

### DIFF
--- a/pkg/controller/plan/adapter/ovirt/builder.go
+++ b/pkg/controller/plan/adapter/ovirt/builder.go
@@ -30,6 +30,7 @@ const (
 	Virtio     = "virtio"
 	Sata       = "sata"
 	Scsi       = "scsi"
+	IDE        = "ide"
 )
 
 // Input types
@@ -411,7 +412,7 @@ func (r *Builder) mapDisks(vm *model.Workload, dataVolumes []cdi.DataVolume, obj
 		switch da.Interface {
 		case VirtioScsi:
 			bus = Scsi
-		case Sata:
+		case Sata, IDE:
 			bus = Sata
 		default:
 			bus = Virtio


ahadas (approved):
commit message - what a nice phrasing :)

mansam (approved):
LGTM

michalskrivanek:
I would prefer a more descriptive commit message - this is specifically for i440fx conversions to q35. IDE disk will most likely not work as SATA as the enumeration is different, but 1) we have to map it somehow and 2) it's still a better shot than virtio somehow.

mansam:
@michalskrivanek Being able to map everything to virtio was the rationale behind using virt-v2v to do guest conversion on RHV VMs. Maybe this and the PR to skip guest conversion need to be reverted until a more holistic set of fixes can be applied at once.

ahadas:
@michalskrivanek see where your message took this :) of course IDE to SATA is only part of i440fx to q35 as IDE is not supported on q35 and RHV does exactly when it converts VMs to q35 (here it's actually easier in this context since we don't preserve the device names, device addresses and such)

ahadas:
> @michalskrivanek Being able to map everything to virtio was the rationale behind using virt-v2v to do guest conversion on RHV VMs. Maybe this and the PR to skip guest conversion need to be reverted until a complete solution can applied at once.

the cost of that is really too high - the extra overhead of executing virt-v2v, the fact it limits the conversion to operating systems that are supported by virt-v2v, the fact it abuses virt-v2v to something it was never intended to (virt-v2v was designed and only tested for migrations from foreign hypervisors to kvm, the way it's used by forklift is one of the things in the list that virt-v2v's maintainer asked us to change..)

mansam:
Got it, thanks @ahadas

PR #453, opened 2022-08-03
Move the end of Reconcile to the defer

mnecas:
Issue: When we set the new condition of status but the controller reconcile
fails before the update of the CRD provider the condition is not saved.

Suggested fix: Add the update of the provider to the defer function which will be
called at the end of reconciliation.


jortel (changes requested):
As a policy, if the reconcile fails, the controller should not update the CR status.  The status could be corrupted and/or reflect incomplete, inaccurate or incoherent status.
The natural remedy is for the event to re-queue and re-reconcile.

PR #454, opened 2022-08-07
MTV-337: Allow insecure connection to Ovirt server

yaacov:
Issue:
We would like to add a "Verify server's SSL certificate" field to the forklift UI, as suggested here: https://github.com/konveyor/forklift-ui/pull/977#issuecomment-1206617874

In some cases it's beneficial for users who do not need this extra security to skip the SSL verification. 

Suggestion:
Add a new `insecureSkipVerify` flag to the oVIrt connection secret to allow insecure connection for oVirt servers.
If this flag is set to `true`, `cacert` field is ignored and an insecure connection is created.

  - [x] update plan client to allow insecure connection
  - [x] update provider connection to allow insecure connection
  - [x] update verification checks
  - [x] add a condition to inform users the provider is using insecure connection 

Ref: https://github.com/konveyor/forklift-ui/pull/977

Signed-off-by: yzamir <yzamir@redhat.com>

### DIFF
--- a/pkg/controller/plan/adapter/ovirt/client.go
+++ b/pkg/controller/plan/adapter/ovirt/client.go
@@ -7,6 +7,7 @@ import (
 	planapi "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/plan"
 	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/ref"
 	plancontext "github.com/konveyor/forklift-controller/pkg/controller/plan/context"
+	"github.com/konveyor/forklift-controller/pkg/controller/provider/container/ovirt"
 	model "github.com/konveyor/forklift-controller/pkg/controller/provider/web/ovirt"
 	ovirtsdk "github.com/ovirt/go-ovirt"
 	cdi "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
@@ -289,6 +290,7 @@ func (r *Client) connect() (err error) {
 		Username(r.user()).
 		Password(r.password()).
 		CACert(r.cacert()).
+		Insecure(ovirt.GetInsecureSkipVerifyFlag(r.Source.Secret)).
 		Build()
 	if err != nil {
 		err = liberr.Wrap(err)
--- a/pkg/controller/provider/validation.go
+++ b/pkg/controller/provider/validation.go
@@ -10,6 +10,7 @@ import (
 	libref "github.com/konveyor/controller/pkg/ref"
 	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
 	"github.com/konveyor/forklift-controller/pkg/controller/provider/container"
+	"github.com/konveyor/forklift-controller/pkg/controller/provider/container/ovirt"
 	core "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -27,6 +28,7 @@ const (
 	ConnectionTestFailed    = "ConnectionTestFailed"
 	InventoryCreated        = "InventoryCreated"
 	LoadInventory           = "LoadInventory"
+	ConnectionInsecure      = "ConnectionInsecure"
 )
 
 //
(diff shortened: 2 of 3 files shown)


yaacov:
@sjd78 hi, this may help with https://github.com/konveyor/forklift-ui/pull/977 please review

pkg/controller/provider/validation.go
```diff
@@ -317,3 +331,38 @@ func (r *Reconciler) inventoryCreated(provider *api.Provider) error {
...
+
+	if insecure {
+		insecureSkipVerifyCnd := libcnd.Condition{
+			Type:     ConnectionInsecure,
+			Status:   True,
+			Reason:   InsecureSkipVerify,
+			Category: Warn,
+			Message:  "TLS is susceptible to machine-in-the-middle attacks unless custom verification is used.",
+		}
+		provider.Status.SetCondition(insecureSkipVerifyCnd)
```
mnecas, line 353: Please rewrite it to, you are not using the `insecureSkipVerifyCnd` anywhere else.
```go
provider.Status.SetCondition(
		libcnd.Condition{
			Type:     ConnectionInsecure,
			Status:   True,
			Reason:   InsecureSkipVerify,
			Category: Warn,
			Message:  "TLS is susceptible to machine-in-the-middle attacks unless custom verification is used.",
		}
)
```

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -316,3 +323,14 @@ func (r *Client) cacert() []byte {
 	}
 	return nil
 }
+
+//
+// Get the insecureSkipVerify boolean value from the ovirt connection secret.
+func (r *Client) insecureSkipVerify() (bool, error) {
```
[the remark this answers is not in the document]
bennyz: > You are not returning errors at any point of the function.

strconv.ParseBool returns (bool, error)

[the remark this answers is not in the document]
mnecas: ahhh in that case I would do it as in the settings
https://github.com/konveyor/forklift-controller/blob/367d2e9e2378ba7419881aff80fd32472a4b8538/pkg/settings/settings.go#L96-L106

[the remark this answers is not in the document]
yaacov: > You are not returning errors at any point of the function.

yes, strconv.ParseBool returns (bool, error)

> Instead of having the same function defined in 2 places maybe move it to some shared lib like https://github.com/konveyor/controller

I'm not sure it's a functionality we neen in `konveyot/controller`, maybe we have a better place to store small common functionalities in this repo, @mansam WDYT ?

pkg/controller/provider/validation.go
yaacov: yes, makes sense to me too, using your suggestion

pkg/controller/plan/adapter/ovirt/client.go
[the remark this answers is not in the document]
yaacov: > Instead of having the same function defined in 2 places maybe move it to some shared lib 

@mnecas I refactored the 3 `insecureSkipVerify` methods into the ovirt pkg, PTAL

yaacov:
@mnecas 
a - refactored the dupllication code to be in the ovirt pkg.
b - refactored the ParseBool err handling to be in-method
please re-review

mnecas (approved):
lgtm

yaacov:
squashing the 7 commits

ahadas (approved):

yaacov:
@mansam please review

pkg/controller/provider/validation.go
```diff
@@ -42,14 +44,15 @@ const (
...
-	Started      = "Started"
+	NotSet             = "NotSet"
+	NotFound           = "NotFound"
+	NotSupported       = "NotSupported"
+	DataErr            = "DataErr"
+	Malformed          = "Malformed"
+	Completed          = "Completed"
+	Tested             = "Tested"
+	Started            = "Started"
+	InsecureSkipVerify = "insecureSkipVerify"
```
mansam, line 55: I think "SkipTLSVerification" for the condition reason instead of "InsecureSkipVerify" would be clearer.

pkg/controller/provider/validation.go
```diff
@@ -202,10 +205,25 @@ func (r *Reconciler) validateSecret(provider *api.Provider) (secret *core.Secret
...
-			"user",
-			"password",
-			"cacert",
+		if ovirt.GetInsecureSkipVerifyFlag(secret) {
+			provider.Status.SetCondition(libcnd.Condition{
+				Type:     ConnectionInsecure,
+				Status:   True,
+				Reason:   InsecureSkipVerify,
+				Category: Warn,
+				Message:  "TLS is susceptible to machine-in-the-middle attacks unless custom verification is used.",
```
mansam, line 214: Maybe "TLS is susceptible to machine-in-the-middle attacks when certificate verification is skipped."

mansam (changes requested):
The code looks good to me, but I have a few suggestions about messages.

pkg/controller/provider/validation.go
yaacov: makes sense to me too, changed

pkg/controller/provider/validation.go
yaacov: The reason for choosing `InsecureSkipVerify` was to keep the name used by the client, but I see the point here as it is a user facing string, makes sense to make it clearer to a user reading the condition, changing as suggested

yaacov:
@mansam please re-review

mansam (approved):
LGTM

mansam:
@yaacov thanks!

PR #455, opened 2022-08-07
Feature page about LUNs migration

liranr23:
This patch introduce the direct LUNs migration. The feature page talks
about the design and implementation of it.

### DIFF
--- a/docs/devel/luns.md
+++ b/docs/devel/luns.md
@@ -1,0 +1,93 @@
+Direct LUNs migration
+==============
+
+Summary
+-------
+
+Usually a VM contains disks. Some of them might be connected as a direct LUN.
+When a user wish to use the LUN disks from another enviroment, he needs them attached to the VM.
+The ISCSI storage where the LUNs reside, have to be reachable to the VM environments.
+The LUN disk should be used in a single place, or else, it will cause a data corruption.
+
+CDI doesn't support LUNs to that matter, we also don't require copying the data, we just need to re-use the same LUN.
+
+Currently, the feature page intend to oVirt only.
+
+Design
+------
+
+The requirements for the LUN migration code are (not in priority order)
+
+1. Gather the LUN information from the source provider.
+2. To ease on the user - less manual actions, more automated migration.
+
+Requirements
+----------------------------------------
+
+1. Destination (OCP cluster) reachability to the ISCSI storage, where the LUN is.
+2. RBAC permission to create `PersistentVolumes` by the controller.
+   - Will be added to the operator manifest.
+
+Risks and Mitigations
+----------------------------------------
+
+1. A VM that will be used while or after the migration with a direct LUN connected.
+   - A warning should be shown before starting the migration.
+   - Detaching the LUN from source VM will split into two cases, based on the migration type:
+     - Cold migration: The LUN disk will be detached before initiating the copy disk. If the migration will fail, re-attach it to the source VM.
+     - Warm migration: Once the migration ends and we shut down the VM on source.
+   - (Optional) UI checkbox to the user, before starting the migration.
+2. An oVirt version with unsupported API. The [PR](https://github.com/oVirt/ovirt-engine/pull/559) provides LUN details is required.
+   - We may check for missing LUN details and block the migration.
+   - (Optional) Have a mechanism to check the oVirt version for future work.
+3. On non-local destination, the controller needs permission to create the `PersistentVolumes`.
+   - It may be enough to be added to the forklift-operator.
+
+Implementation
+----------------------------------------
+
+The controller gather data from source provider and saves them to the inventory.
+As part of the data gathering we will gather the LUN disks data. We have the VMs disk attachments where we follow to the disks.
+Within the disks API, if the _storage\_type_ is `LUN`, we will gather from the API endpoint `api/disks/<disk_id>` the size, address, IQN and the mapping of the LUN.
+Those will be found withing the `logical_unit` element of the `lun_storage`.
+On the migration plan a new steps will be added, creating a `PersistentVolume`(PV) and a `PersistentVolumeClaim`(PVC) per LUN disk.
+Both of them will contain the relevant data to use the LUN, the PVC will be bound the the PV by a unique label, `<vm_id>-<vm_name>-vol-<volume-number>`.
+When the VM specification created, the controller will add a LUN disk using the PVC that was created as [described by kubevirt](https://kubevirt.io/user-guide/virtual_machines/disks_and_volumes/#lun).
+
+Relevant validations will be performed to know we have the LUN details:
+
+- Allow `LUN` storage type to pass the validation - until know only `image` type is allowed to migrate.
+- LUN details exist on the source provider API, the data is neccassary in order to migrate.
+- SCSI persistent reservation (exists), may required further warning when LUN is detected.
+- (Optional) Make sure the VM is in `down` on source provider (might be not required and be part of the execution plan).
+- In case the storage requires authentication, we may need to add secrets.
+
+Notes:
+
+- CSI is optional, it won't able to select the relevant LUN we wish to use. However, we can add the relevant `StorageClass` of the CSI to the `PVC` specifications.
+  - Will require creating a mapping for each LUN to the relevant `StorageClass`.
+  - User will need to pre-install the CSI and have the `StorageClass`.
+  - `PV` creation still needed to have the specific LUN details.
+
+Execution Flow
+----------------------------------------
+
+1. The inventory will be filled based on the provider.
+   - On each disk, try to gather `lun_storage` data.
+2. Validation rules will be executed.
+3. Two steps will be added to the migration:
+   - `CreatePersistentVolumes`, which will create the relevant `PV` based on the `LUN` details.
+   - `CreatePersistentVolumeClaims`, which will create the relevant `PVC` and **bound** to the above `PV`.
+     - Might be possible to skip and set into the VM specification with all the relevant `PVC` details.
+On each step, it first will check the existence of the CRs, when they won't be available, it will be created.
+4. When building the VM specifications, create the `LUN` disk and volume associated with the `PVC`.
+5. Once the migration ends successfully, detach the `LUN` disk from the source VM.
+   - Cold migration: The LUN disk will be detached before initiating the copy disk.
+   - Warm migration: Once the migration ends and we shut down the VM on source.
+6. If the migration failed and it archieved, remove any new CR that was created (PV/PVC).
+   - In cold migration, re-attach the LUN on the source VM.
+
+Future
+----------------------------------------
+
+- Support SCSI persistent reservation once [kubevirt PR](https://github.com/kubevirt/kubevirt/pull/8210) will be merged.


docs/devel/luns.md
```diff
@@ -0,0 +1,56 @@
...
+----------------------------------------
+
+1. A LUN being used by a VM the user wish to migrate, without reachability to the destination.
+2. A VM that will be used while or after the migration with a direct LUN connected.
+   - A warning should be shown before starting the migration.
+   - (Optional) UI checkbox to the user, before starting the migration.
+3. An oVirt version with unsupported API. The [PR](https://github.com/oVirt/ovirt-engine/pull/559) provides LUN details is required.
+   - We may check for missing LUN details and skip the disk migration.
+   - (Optional) Have a mechanism to check the oVirt version for future work.
+4. Creating a `PersistentVolume` requires the file system type: `ext4` or `xfs` of the LUN, which we don't know.
```
bennyz, line 38: we can't use `volumeMode: Block`?

docs/devel/luns.md
```diff
@@ -0,0 +1,56 @@
...
+
+1. Destination (OCP cluster) reachability to the ISCSI storage, where the LUN is.
+2. RBAC permission to create `PersistentVolumes` by the controller.
+
+Risks and Mitigations
+----------------------------------------
+
+1. A LUN being used by a VM the user wish to migrate, without reachability to the destination.
+2. A VM that will be used while or after the migration with a direct LUN connected.
+   - A warning should be shown before starting the migration.
```
michalskrivanek, line 33: we need to block starting of the original VM in some way. We can't risk corruption

liranr23: We can't really own that VM on source. We can detach the LUN from the VM. WDYT?

docs/devel/luns.md
liranr23: Both required, the `volumeMode: Block` is hard-coded for LUN, but we do have `fsType` spec under the `iscsi`. By default it would be `ext4` (when not provided).

michalskrivanek: isnt't that only for volumemount? if it's block it shoulnd't need any fs.

docs/devel/luns.md
```diff
@@ -0,0 +1,56 @@
...
+==============
+
+Summary
+-------
+
+VM contains disks. Some of them might be connected as a direct LUN.
+When a user wish to re-use the LUN disks, he needs them attached to the VM.
+The LUN should be used in a single place, or else, it will cause a data corruption.
+The storage where the LUNs reside, needs to be reachable to the VM environment.
+This feature page mostly about import from oVirt.
```
ahadas, line 11: as far as I know we never supported importing LUNs from vsphere (definitely not when importing through RHV, most probably also not from IMS) so let's scope this only to oVirt/RHV

docs/devel/luns.md
```diff
@@ -0,0 +1,56 @@
...
+
+Design
+------
+
+The requirements for the LUN migration code are (not in priority order)
+
+1. Gather the LUN informnation from the source destination.
+2. To ease on the user - less manual actions, more automated migration.
+3. Save unnecessary risks.
```
ahadas, line 20: you probably meant to avoid risks, not to save them, right? :)

docs/devel/luns.md
```diff
@@ -0,0 +1,56 @@
...
+
+Requirements
+----------------------------------------
+
+1. Destination (OCP cluster) reachability to the ISCSI storage, where the LUN is.
+2. RBAC permission to create `PersistentVolumes` by the controller.
```
ahadas, line 26: that's something we'll need to take care of if we need this

docs/devel/luns.md
michalskrivanek: well - we have to modify it in this case. Not sure if it's worth to provide an option for not detaching it. Maybe for development/testing

docs/devel/luns.md
```diff
@@ -0,0 +1,56 @@
...
+
+Risks and Mitigations
+----------------------------------------
+
+1. A LUN being used by a VM the user wish to migrate, without reachability to the destination.
```
ahadas, line 31: I wouldn't say it's a "risk" - I would expect the PVC to stay in pending mode in this case

docs/devel/luns.md
ahadas: +1 for detaching it at the end of the migration, right before the vm on the destination cluster can be started

docs/devel/luns.md
```diff
@@ -0,0 +1,56 @@
...
+Implementation
+----------------------------------------
+
+The controller gather data from source provider and saves them to the inventory.
+As part of the data gathering we will gather the LUN disks data. We have the VMs disk attachments where we follow to the disks.
+Within the disks API, if the _storage\_type_ is `LUN`, we will gather from the API endpoint `api/disks/<disk_id>` the size, address, IQN and the mapping of the LUN.
+Those will be found withing the `logical_unit` element of the `lun_storage`.
+On the migration plan a new steps will be added, creating a `PersistentVolume`(PV) and a `PersistentVolumeClaim`(PVC) per LUN disk.
+Both of them will contain the relevant data to use the LUN, the PVC will be bound the the PV by a unique label, `<vm_name>-vol-<volume-number>`.
+When the VM specification created, the controller will add a LUN disk using the PVC that was created as [described by kubevirt](https://kubevirt.io/user-guide/virtual_machines/disks_and_volumes/#lun).
```
ahadas, line 51: did you check if cdi supports defining luns? if it's not, wouldn't it make sense to add it there?

liranr23: all i could find is that CDI can create PVC. Which feels more effort as I need to put all the details there(as I need to the PVC). https://github.com/kubevirt/containerized-data-importer/blob/main/doc/datavolumes.md#target-storagepvc

docs/devel/luns.md
liranr23: i was based on https://docs.openshift.com/container-platform/4.6/storage/persistent_storage/persistent-storage-iscsi.html , but reading carefully https://kubernetes.io/docs/concepts/storage/persistent-volumes/#volume-mode, maybe it is possible to drop.

docs/devel/luns.md
ahadas: why would it require more efforts? wouldn't a data volume with 'source' that contains all the data you put into the PV be enough? Benny raised a valid point earlier today that if there's a cdi of the same storage type that we use in ovirt (e.g., netapp) it would make sense to use it here so that would be just another storage class that we specify on the pvc I suppose

liranr23: After discussing with Alexander Wels, CDI doesn't support it as I inspected. But he pointed out [this doc](https://github.com/kubevirt/kubevirt/blob/main/docs/iscsi-authentication.md) and his own words about the differences of my suggested approach and the doc above:

> Well that is slightly different, that takes a disk.img (or block volume) in a PVC and exposes it as a LUN inside the VM. It might not actually exist on an iSCSI device. It just looks like that to the VM. The example I pointed to is actually connecting to the iSCSI device and using the provided information to find the LUN on the device. You will notice that we need to provide an auth secret so the VM can connect to the device.
> Now it is possible to create a PV that points to an iSCSI device and then create a PVC that binds to the PV, and then use that PVC in the VM, and we can expose that PVC as a lun inside the VM, but that adds a bunch of extra layers vs just pointing directly to the LUN

ahadas: it's not that clear where that flow of exposing disk.img to a LUN inside the VM came from - that message is a bit out of context. but ok, that makes sense - let's try to do it that way. I also noticed that LUN is not part of the sources supported by CDI, but I was wondering if it would make sense to add it there rather than creating the PV and correlating it to the VM separately. if there's a way to do it in a k8s way, it's even better - let's check it

alicefr: CDI relies on `qemu-img` for converting the input VM image and only supports importing VM images to block disks or raw images. KubeVirt already supports direct LUNs but to my knowledge, we don't have a way to populate a direct LUN. We might be able to extend CDI to understand LUNs, but this will require some research

docs/devel/luns.md
```diff
@@ -0,0 +1,54 @@
+Direct LUNs migration
+==============
+
+Summary
+-------
+
+VM contains disks. Some of them might be connected as a direct LUN.
+When a user wish to re-use the LUN disks, he needs them attached to the VM.
+The LUN should be used in a single place, or else, it will cause a data corruption.
```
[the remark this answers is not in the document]
ahadas: I think it would be ok to issue a warning when there's no scsi reservation configured (on the oVirt side) - so the user would know we can't protect from this. I don't think we can do much in this case..

docs/devel/luns.md
```diff
@@ -0,0 +1,54 @@
...
+The storage where the LUNs reside, needs to be reachable to the VM environment.
+This feature intend to support oVirt only.
+
+Design
+------
+
+The requirements for the LUN migration code are (not in priority order)
+
+1. Gather the LUN informnation from the source destination.
```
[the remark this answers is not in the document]
ahadas:
```suggestion
1. Gather the LUN information from the source destination.
```

docs/devel/luns.md
liranr23: Thanks Alice!

docs/devel/luns.md
liranr23: Sometimes the VM may start before the migration ends. We need to be delicate here.

docs/devel/luns.md
liranr23: +1 on the operator manifest

docs/devel/luns.md
ahadas: yeah, the vm might start on oVirt during the migration but as long as it can't start on Openshift Virtualization, we're safe, and it shouldn't be possible to initiate a VMI until the data volumes are ready - so if we can detach the LUN on the source at the end of the migration, before switching the data volumes to ready, it would be fine

docs/devel/luns.md
```diff
@@ -0,0 +1,82 @@
...
+   - Will be added to the operator manifest.
+
+Risks and Mitigations
+----------------------------------------
+
+1. A VM that will be used while or after the migration with a direct LUN connected.
+   - A warning should be shown before starting the migration.
+   - Detaching the LUN from source VM, once the migration completes.
+   - (Optional) UI checkbox to the user, before starting the migration.
+   - The migration _must_ be a cold migration.
```
ahadas, line 36: why does it must to be a cold migration? it should be easier to support warm migration when you assume access to the storage from both places

docs/devel/luns.md
```diff
@@ -0,0 +1,82 @@
...
+   - It may be enough to be added to the forklift-operator.
+
+Implementation
+----------------------------------------
+
+The controller gather data from source provider and saves them to the inventory.
+As part of the data gathering we will gather the LUN disks data. We have the VMs disk attachments where we follow to the disks.
+Within the disks API, if the _storage\_type_ is `LUN`, we will gather from the API endpoint `api/disks/<disk_id>` the size, address, IQN and the mapping of the LUN.
+Those will be found withing the `logical_unit` element of the `lun_storage`.
+On the migration plan a new steps will be added, creating a `PersistentVolume`(PV) and a `PersistentVolumeClaim`(PVC) per LUN disk.
```
ahadas, line 52: why do we need to create a step for creating a PVC (as opposed to posting it on the VM itself)?

docs/devel/luns.md
```diff
@@ -0,0 +1,82 @@
...
+Both of them will contain the relevant data to use the LUN, the PVC will be bound the the PV by a unique label, `<vm_id>-<vm_name>-vol-<volume-number>`.
+When the VM specification created, the controller will add a LUN disk using the PVC that was created as [described by kubevirt](https://kubevirt.io/user-guide/virtual_machines/disks_and_volumes/#lun).
+
+Relevant validations will be performed to know we have the LUN details:
+
+- Allow `LUN` storage type on existing validation.
```
ahadas, line 58: what does this mean?

docs/devel/luns.md
```diff
@@ -0,0 +1,82 @@
...
+- LUN details exist on the provider.
```
ahadas, line 59: on the provider? you mean on the disk we fetched from the source system?

docs/devel/luns.md
```diff
@@ -0,0 +1,82 @@
...
+- SCSI persistent reservation (exists), may required further warning when LUN is detected.
+- (Optional) Make sure the VM is in `down` on source provider(might be not required and be part of the execution plan).
```
ahadas, line 61: seems related to the part above which limits this to cold migrations, I can't see why

docs/devel/luns.md
```diff
@@ -0,0 +1,82 @@
...
+
+Execution Flow
+----------------------------------------
+
+1. The inventory will be filled based on the provider.
+   - On each disk, try to gather `lun_storage` data.
+2. Validation rules will be executed.
+3. Two steps will be added to the migration:
+   - `CreatePersistentVolumes`, which will create the relevant `PV` based on the `LUN` details.
```
ahadas, line 70: did you consider checking on the destination cluster if we have a storage provisioner that can prepare the PV for us first?

docs/devel/luns.md
```diff
@@ -0,0 +1,82 @@
...
+   - `CreatePersistentVolumeClaims`, which will create the relevant `PVC` and **bound** to the above `PV`.
+On each step, it first will check the existence of the CRs, when they won't be available, it will be created.
+4. When building the VM specifications, create the `LUN` disk and volume associated with the `PVC`.
+5. Once the migration ends successfully, detach the `LUN` disk from the source VM.
```
ahadas, line 74: that might be too late - if all the disks were migrated and the volumes are ready, one can start the vm on the destination cluster
we need to be more specific here and explain how we make sure the vm cannot be started before we detach the lun from the source?

docs/devel/luns.md
```diff
@@ -0,0 +1,82 @@
...
+6. If the migration failed and it archieved, remove any new CR that was created (PV/PVC).
```
ahadas, line 75: and also attach the lun back to the vm on the source system?

and what failures do you refer to here? let's say that the destination cluster doesn't have access to the lun (btw, there's no preliminary check for that, right?) - would we still detach the lun from the source? the downside is that the user may end up in a vm that cannot be started on the destination without further changes, and cannot start the vm on the source without attaching the lun back

docs/devel/luns.md
liranr23: because the LUN detachment, i meant detaching from the VM. AFAIK it's not possible when the VM is up.

docs/devel/luns.md
liranr23: you mean that the PVC will be created based on the VM spec? if so, how it will know which PV to bound?

docs/devel/luns.md
liranr23: i will elaborate on the next push. i meant that now the validation let you migrate only VMs that have disks of type `image`. I want to let `lun` as well.

docs/devel/luns.md
liranr23: yes, to make sure it contains the additional API from benny, to have the details to migrate the disk.

docs/devel/luns.md
liranr23: explained on the other part.

docs/devel/luns.md
liranr23: no, can you elaborate? i did try on my cluster and i had no permission to do it out of the box.

docs/devel/luns.md
ahadas: we anyway shut down the vm during the migration, so once it's off and you do the final copy of other disks (if exist), you can detach the lun

docs/devel/luns.md
liranr23: no, the idea was to detach the disk only when the migration succeed, until then it will be available on the source.
the failures i refer to is each problem that may occur in the migration process.
we don't check access to the lun on destination cluster. you are right about having a "succeed" migration when the destination have no access and we will remain bald from both sides (until manual action is executed).
a warning / another step of validation to check access + documentation?

docs/devel/luns.md
ahadas: > you mean that the PVC will be created based on the VM spec?

yes, the pvc would be part of the vm specification (as the disks are today) and the pvc would be derived from that

> if so, how it will know which PV to bound?

the same way as you planned - by configuring the selection to match the pv

docs/devel/luns.md
ahadas: ok, so please rephrase - it's not on the provider then..

docs/devel/luns.md
ahadas: let's say that the user uses netapp on rhv and there's a netapp provisioner on the target ocp cluster - I'd expect to delegate the configuration of the lun to the netapp provisioner rather than posting such "static" PVs. posting PVs should really be done as a last resort, if there is no other way to get it

docs/devel/luns.md
```diff
@@ -0,0 +1,82 @@
...
+This feature intend to support oVirt only.
+
+Design
+------
+
+The requirements for the LUN migration code are (not in priority order)
+
+1. Gather the LUN information from the source destination.
+2. To ease on the user - less manual actions, more automated migration.
+3. Avoid unnecessary risks.
```
michalskrivanek, line 20: I guess this section can be dropped

docs/devel/luns.md
```diff
@@ -0,0 +1,82 @@
...
+Risks and Mitigations
+----------------------------------------
+
+1. A VM that will be used while or after the migration with a direct LUN connected.
+   - A warning should be shown before starting the migration.
+   - Detaching the LUN from source VM, once the migration completes.
+   - (Optional) UI checkbox to the user, before starting the migration.
+   - The migration _must_ be a cold migration.
+2. An oVirt version with unsupported API. The [PR](https://github.com/oVirt/ovirt-engine/pull/559) provides LUN details is required.
+   - We may check for missing LUN details and skip the disk migration.
```
michalskrivanek, line 38: it's fine to block

docs/devel/luns.md
```diff
@@ -0,0 +1,82 @@
...
+   - (Optional) Have a mechanism to check the oVirt version for future work.
+3. Creating a `PersistentVolume` requires the file system type: `ext4` or `xfs` of the LUN, which we don't know.
+   - Might not be required for `Block` type.
```
michalskrivanek, line 41: yes, this is imho wrong, you want "block" on the target, not a LUN with fs

docs/devel/luns.md
```diff
@@ -0,0 +1,82 @@
...
+4. On non-local destination, the controller needs permission to create the `PersistentVolumes`.
+   - It may be enough to be added to the forklift-operator.
```
michalskrivanek, line 43: how would it work? 
maybe we just need to check for this permission and ask user to add it?

docs/devel/luns.md
```diff
@@ -0,0 +1,82 @@
+Direct LUNs migration
+==============
+
+Summary
+-------
+
+VM contains disks. Some of them might be connected as a direct LUN.
+When a user wish to re-use the LUN disks, he needs them attached to the VM.
+The LUN should be used in a single place, or else, it will cause a data corruption.
+The storage where the LUNs reside, needs to be reachable to the VM environment.
```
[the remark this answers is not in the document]
liranr23: better now?

docs/devel/luns.md
```diff
@@ -0,0 +1,82 @@
...
+This feature intend to support oVirt only.
```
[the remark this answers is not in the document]
liranr23: ok, i changed it that the FP talks about ovirt.

docs/devel/luns.md
```diff
@@ -0,0 +1,82 @@
...
+
+Design
+------
+
+The requirements for the LUN migration code are (not in priority order)
+
+1. Gather the LUN information from the source destination.
```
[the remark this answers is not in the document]
liranr23: done

docs/devel/luns.md
liranr23: i think it might be a hard catch.
the importer executing the copyDisk. we check when it's done. I think when it is - the VM might be available to use even when the migration didn't complete. the reconcile will see when it's done and pass to the next phase. overall i think if the user manually "on it" he may have a few seconds to start the vm before the migration of that vm is done in mtv.

docs/devel/luns.md
liranr23: TBH, i'm really not sure about it. adding it to the manifest in the operator should be enough when MTV is in the same cluster as the destination. the controller does have permissions to create CRs on destination cluster. If so, this is the place where we say what permission are granted to the controller. but only testing it will give the answer (or if anyone knows).

docs/devel/luns.md
liranr23: done

docs/devel/luns.md
liranr23: done

docs/devel/luns.md
liranr23: I need to test it but sure, if it's possible.. why not.

docs/devel/luns.md
liranr23: We did talk more about it offline - currently the next step is to check if CSI can do the job and save us from creating the PV.
In that case, we will need to check the vendor of the LUN (provided in the API) - automating the storage class, or:
create mapping like the current SD mapping, just based on LUN to Storage Class with the relevant CSI to use.
Then, it may seem we really don't need much other than setting up the PVC with the mapped SC. i am working to check this flow.
Once i'll see it works, i'll refresh the FP.

docs/devel/luns.md
ahadas: if there's no "hook" to plug this to CDI before switching the data volumes to 'ready' state, I'd do it once the vm on the source is down during warm migration and at the beginning of cold-migration

PR #456, opened 2022-08-09
[WIP] migrating lun disks from ovirt

liranr23:
WIP - Migrate LUN disks from oVirt provider.

Missing - checking the existence, cleaning when archive, gathering LUN data from source and more
as pointed out in the feature page.


ahadas:
let's move this to kubev2v/forklift as well

PR #457, opened 2022-08-09
Change code structure

ahadas:
The current state of forklift is not so developer-friendly since the code is scattered across various repos in the konveyor organization so developers might need to make code changes in various repos for a single feature, and to get code from several places to set up an environment for automated end-to-end tests (CI).

Currently, this PR proposes two changes:
(1) To vendor the dependencies of the controller
(2) To use bazel for building the controller binary

The idea though is broader - to have a single repo that includes all (or most of) the components of forklift, much like kubevirt/kubevirt.
That way we can, for example, introduce forklift-api pod if we want to more easily (no need to create a separate repo)
The code within the repo can be owned by different people using OWNERS files for each component/package


fbladilo:
@ahadas Please be careful here, new build systems are designed to track and build changes per component in individual SCM repos, micro services projects around k8s tend to be setup this way as it is extremely flexible (among various other reasons). Going for a monolithic approach will limit your options in this sense and perhaps force you to use/develop your own tooling. In other words, it makes sense to weight the developer friendliness weight of the decision vs the cost. My 2 cents.

ahadas:
> @ahadas Please be careful here, new build systems are designed to track and build changes per component in individual SCM repos, micro services projects around k8s tend to be setup this way as it is extremely flexible (among various other reasons). Going for a monolithic approach will limit your options in this sense and could potentially force you to use/develop your own tooling (this is seen in some very large scale projects). In other words, it makes sense to measure the developer friendliness weight of the decision vs the _true_ cost. My 2 cents.

oh sorry @fbladilo , I missed your comment for some reason (I didn't expect someone to review this draft PR yet :) )
ack, yeah - the main motivation for this is twofold:
1. to be able to post patches that cover both validations and logic within the controller in a single PR
2. more importantly, to be able to clone a single repo and deploy all relevant things from it for end-to-end tests

you're right that it needs to take productization aspect into account (looking forward to your session on that ;) ), I believe there should be good tooling for that because [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt) is structured in a similar way, but yeah - productization would definitely need to be considered before getting this in

PR #458, opened 2022-08-16
ovirt: perform snapshot removal sequentially

bennyz:
Perform snapshot removals sequentially, as the first snapshot removal will block the others and will result in only one snapshot being removed

Signed-off-by: Benny Zlotnik <bzlotnik@redhat.com>


pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -64,16 +65,44 @@ func (r *Client) RemoveSnapshots(vmRef ref.Ref, precopies []planapi.Precopy) (er
...
-	for i := range precopies {
-		snapService := snapsService.SnapshotService(precopies[i].Snapshot)
-		_, err = snapService.Remove().Async(true).Query("correlation_id", r.Migration.Name).Send()
-		if err != nil {
-			err = liberr.Wrap(err)
+	go func() {
+		for i := range precopies {
+			snapService := snapsService.SnapshotService(precopies[i].Snapshot)
+			correlationID := fmt.Sprintf("precopy_%d_finalize", i)
+			_, err = snapService.Remove().Query("correlation_id", correlationID).Send()
```
ahadas, line 72: how about trying several times to initiate the first remove-snapshot when it fails (due to memory lock) as we know it would most probably happen due to image transfer

ahadas:
and please correlate this with https://bugzilla.redhat.com/2053183

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -64,16 +65,44 @@ func (r *Client) RemoveSnapshots(vmRef ref.Ref, precopies []planapi.Precopy) (er
...
 	snapsService := vmService.SnapshotsService()
-	for i := range precopies {
-		snapService := snapsService.SnapshotService(precopies[i].Snapshot)
-		_, err = snapService.Remove().Async(true).Query("correlation_id", r.Migration.Name).Send()
-		if err != nil {
-			err = liberr.Wrap(err)
+	go func() {
+		for i := range precopies {
+			snapService := snapsService.SnapshotService(precopies[i].Snapshot)
+			correlationID := fmt.Sprintf("precopy_%d_finalize", i)
```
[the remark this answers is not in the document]
bennyz: yeah it's temporary, I had some issues with the snapshot id, I didn't look into it yet so I placed this correlation id for now

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -64,16 +65,44 @@ func (r *Client) RemoveSnapshots(vmRef ref.Ref, precopies []planapi.Precopy) (er
...
-	}
+	}()
+
 	return
 }
 
+func allJobsFinished(jobs []*ovirtsdk.Job) bool {
+	for _, job := range jobs {
+		status, _ := job.Status()
+		if status != ovirtsdk.JOBSTATUS_FINISHED && status != ovirtsdk.JOBSTATUS_FAILED {
```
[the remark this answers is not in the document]
bennyz: I'll check again, but IIRC this status is only used for gluster stuff

pkg/controller/plan/adapter/ovirt/client.go
bennyz: You mean checking if it fails generally or fails because of the memory lock? Do we have a reliable way to know it failed because because of a memory lock? Because I'm only aware of parsing the error message which is probably not as reliable

pkg/controller/plan/adapter/ovirt/client.go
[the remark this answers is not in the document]
bennyz: Added the start of the snapshot's id, there is a limit on the correlation id length. I can just put the entire id and that's it, but I think it would make it harder to trace in the logs

bennyz:
@mansam Do you see issues with this approach or is it a sensible solution to cleaning up oVirt's snapshots sequentially?

pkg/controller/plan/adapter/ovirt/client.go
ahadas: we talked about this offline that we don't have a good/reliable way to distinguish between failures to acquire the lock and other failures, but in this particular case it seems reasonable to assume that failure to initiate the snapshot removals are due to image transfer and try several time, unless we have a better way to sync with what's going on in CDI

bennyz: Thought about this some more, image transfer entities have an image id (which is actually the disk id)
```xml
<image_transfer href="/ovirt-engine/api/imagetransfers/e2294cfc-9b41-43eb-b7ef-fcffd15ce9b9" id="e2294cfc-9b41-43eb-b7ef-fcffd15ce9b9">
	  <actions>
	  ...
	  </actions>
	  <active>true</active>
	  <direction>download</direction>
	  <format>raw</format>
	  <inactivity_timeout>60</inactivity_timeout>
	  <phase>finished_success</phase>
	  <shallow>false</shallow>
	  <timeout_policy>cancel</timeout_policy>
	  <transferred>1427243008</transferred>
	  <host href="/ovirt-engine/api/hosts/00509b81-fa25-4953-ba07-824dfc4e32d0" id="00509b81-fa25-4953-ba07-824dfc4e32d0"/>
	  <image id="790bd328-0525-45c5-925e-52f3b4428414"/>
</image_transfer>
``

So I I'll see if simply searching if the disk id is participating in any active transfer is enough

Comment (ahadas) [reply pkg/controller/plan/adapter/ovirt/client.go:74]: we can do something like optimistic locking :) trying to remove it and if that fails confirming that it's due to an ongoing image transfer this way

File: pkg/controller/plan/adapter/ovirt/client.go [hunk 4]
```diff
@@ -318,3 +355,78 @@ func (r *Client) cacert() []byte {
...
+				Jitter:   0,
+				Steps:    3,
+			}
+
+			err = wait.ExponentialBackoff(backoff, func() (bool, error) {
+				snapshotDisksResponse, _ := snapService.DisksService().List().Send()
+				snapshotDisks := snapshotDisksResponse.MustDisks().Slice()
+				for _, disk := range snapshotDisks {
+					if transferred, err := r.isDiskBeingTransferred(disk.MustId()); transferred || err != nil {
+						r.Log.Info("Disk is being transferred, retrying..")
```
bennyz, line 383: Add the number of retries to the log

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -318,3 +355,78 @@ func (r *Client) cacert() []byte {
...
+			// Assume we could not remove because of an ongoing transfer
+			backoff := wait.Backoff{
+				Duration: time.Second * 3,
+				Factor:   0,
+				Jitter:   0,
+				Steps:    3,
+			}
+
+			err = wait.ExponentialBackoff(backoff, func() (bool, error) {
+				snapshotDisksResponse, _ := snapService.DisksService().List().Send()
```
bennyz, line 379: fill error check

ahadas (commented):
the logic seems good to me

ahadas (approved):

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -281,6 +278,46 @@ func (r *Client) getDiskSnapshot(diskID, targetSnapshotID string) (diskSnapshotI
 	return
 }
 
+func (r *Client) getJobs(correlationID string) (ovirtJob []*ovirtsdk.Job, err error) {
+
+	// TODO: figure out a better way
+	r.connect()
```
mansam, line 284: Why is this connecting here?

bennyz: Otherwise the client is not connected and API calls are NPE'd, I believe it's because the connection is closed outside, in migratio.go

mansam: Did you test this with more than one VM on the migration plan? This is not safe.

A single Client is instantiated at the start of `Migration.Run()` (in `Migration.init()`), which is then closed in a defer at the end of `Migration.Run()`. Inside of `Run()`, `Migration.execute()` gets called for each VM migration on the plan that is currently in flight, plus once more if there's an additional migration ready and able to start. If you had only a single VM on the migration plan and the remove snapshot goroutine was long running, it's likely that `Run()` would return and the connection would be closed before the goroutine finished. However, it's possible for there to be many VM migrations running, and all of them use the same client in a given reconcile. These VM migrations can all be at different steps in the process, so it's entirely possible with this PR that your first VM could finish, delete its snapshots, and then call `client.Close()` inside of `removePrecopies()`, which would mean that the client was closed for all of the following calls to `execute()`.

You need to be sure that the goroutine isn't sharing the client with anything else. Since you're kicking off this goroutine from inside of the client, you'll probably need to rethink this somewhat.

bennyz: Thanks Sam, and yes, it is a problem I was aware of, I just wasn't sure how to approach it before and decided to go on with the actual snapshot removal first

anyway, I've updated this PR and made it into a slightly larger change. It feels like the main issue is that because of oVirt's shenanigans using RemoveSnapshots is a bit too high of a denominator because of the need for more control for how it is executed.

So what I went with is to add a new interface method to finalize the plan in (`end()`) where oVirt's client will start the asynchronous sequential snapshot removal and can managed the lifecycle of the connection to the SDK. Now, RemoveSnapshots() will do nothing for oVirt.

I've tested it with multiple VMs and it looks ok, I'm not sure about whether this is the best way to do this or if I placed it correctly, but this is what we came up with trying to both control the lifetime of the connection and not have a new connection created for each VM

pkg/controller/plan/migration.go
```diff
@@ -1175,6 +1179,12 @@ func (r *Migration) updateCopyProgress(vm *plan.VMStatus, step *plan.Step) (err
 		return
 	}
 
+	if !r.Plan.Spec.Warm {
```
mansam, line 1182: Why did you add this here? `updateCopyProgress` is not specific to warm migration, it tracks copy progress for all disk transfers.

pkg/controller/plan/migration.go
```diff
@@ -1331,6 +1341,19 @@ func (r *Migration) setDataVolumeCheckpoints(vm *plan.VMStatus) (err error) {
 	return
 }
 
+func (r *Migration) getPrecopiesMap() map[ref.Ref][]plan.Precopy {
```
mansam, line 1344: There's no need for this. You could just pass the slice of VMStatus structs instead of converting it into a map first.

pkg/controller/plan/migration.go
bennyz: you're right, this should not be here, I had some applied changes I had to work with and accidentally left this, I'll remove it promptly

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -318,3 +347,104 @@ func (r *Client) cacert() []byte {
 	}
 	return nil
 }
+
+func (r Client) Finalize(vmPrecopies map[ref.Ref][]planapi.Precopy) {
+	r.connect()
```
mansam, line 352: Isn't there the same issue here, where since you're launching this in a goroutine that `Run()` could terminate and fire its `defer`, causing the connection to be closed after this starts?

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -318,3 +343,108 @@ func (r *Client) cacert() []byte {
 	}
 	return nil
 }
+
+func (r Client) Finalize(vms []*planapi.VMStatus) {
+	r.connect()
```
bennyz, line 348: @mansam this should operate on it's own copy of the client and will outlive `Run()`, I have tested this with 3 VMs, each with 4 precopies, and some of the precopies were still there after `Run()` finished, and were successfully removed

pkg/controller/plan/adapter/ovirt/client.go
```diff
@@ -281,6 +270,42 @@ func (r *Client) getDiskSnapshot(diskID, targetSnapshotID string) (diskSnapshotI
...
+func (r *Client) getJobs(correlationID string) (ovirtJob []*ovirtsdk.Job, err error) {
+	jobService := r.connection.SystemService().JobsService().List()
+	jobResponse, err := jobService.Search(fmt.Sprintf("correlation_id=%s", correlationID)).Send()
+	if err != nil {
+		err = liberr.Wrap(err)
+		return
+	}
+	ovirtJobs, ok := jobResponse.Jobs()
+	if !ok {
+		err = liberr.New(fmt.Sprintf("Job %s source lookup failed", correlationID))
```
bennyz, line 282: use kvpair instead

pkg/controller/plan/migration.go
bennyz: thanks, done

pkg/controller/plan/adapter/ovirt/client.go
mansam: Yeah, I see that you didn't use a pointer receiver here in the function signature which should avoid that particular issue.

ahadas:
@bennyz let's continue this in kubev2v/forklift

PR #459, opened 2022-08-22
Support ovirt CPU Policy

liranr23:
When importing a VM from oVirt it has the property of CPU Pinning Policy.
Only having `none` or `dedicated` is supported on CNV. Therefore, only both
of them will be respected.
As for `dedicated` the user should enable the `CPU manager` on the nodes,
or else the VM won't be able to schedule and the migration will hang.
NUMA is out of the scope for this PR.
More about the feature at the [dedicated cpu resources page](https://kubevirt.io/user-guide/virtual_machines/dedicated_cpu_resources/).

### DIFF
--- a/pkg/controller/plan/adapter/ovirt/builder.go
+++ b/pkg/controller/plan/adapter/ovirt/builder.go
@@ -2,6 +2,9 @@ package ovirt
 
 import (
 	"fmt"
+	"path"
+	"strings"
+
 	liberr "github.com/konveyor/controller/pkg/error"
 	libitr "github.com/konveyor/controller/pkg/itinerary"
 	"github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1/plan"
@@ -14,8 +17,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	cnv "kubevirt.io/client-go/api/v1"
 	cdi "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
-	"path"
-	"strings"
 )
 
 // BIOS types
@@ -56,6 +57,15 @@ const (
 	DefaultWindows = "win10"
 	DefaultLinux   = "rhel8.1"
 	Unknown        = "unknown"
+)
+
+// CPU Pinning Policies
+const (
+	None           = "none"
+	Manual         = "manual"
+	ResizeAndPin   = "resize_and_pin_numa"
+	Dedicated      = "dedicated"
+	IsolateThreads = "isolate_threads"
 )
 
 // Map of ovirt guest ids to osinfo ids.
@@ -358,6 +368,10 @@ func (r *Builder) mapCPU(vm *model.Workload, object *cnv.VirtualMachineSpec) {
 		Cores:   uint32(vm.CpuCores),
 		Threads: uint32(vm.CpuThreads),
 	}
+	if vm.CpuPinningPolicy == Dedicated {
+		object.Template.Spec.Domain.CPU.DedicatedCPUPlacement = true
+	}
+
 }
 
 func (r *Builder) mapFirmware(vm *model.Workload, cluster *model.Cluster, object *cnv.VirtualMachineSpec) {
--- a/pkg/controller/provider/model/ovirt/model.go
+++ b/pkg/controller/provider/model/ovirt/model.go
@@ -131,6 +131,7 @@ type VM struct {
 	CpuCores                    int16            `sql:""`
 	CpuThreads                  int16            `sql:""`
 	CpuAffinity                 []CpuPinning     `sql:""`
+	CpuPinningPolicy            string           `sql:""`
 	CpuShares                   int16            `sql:""`
 	Memory                      int64            `sql:""`
 	BalloonedMemory             bool             `sql:""`
--- a/pkg/controller/provider/container/ovirt/resource.go
+++ b/pkg/controller/provider/container/ovirt/resource.go
@@ -1,9 +1,10 @@
 package ovirt
 
 import (
-	model "github.com/konveyor/forklift-controller/pkg/controller/provider/model/ovirt"
 	"sort"
 	"strconv"
+
+	model "github.com/konveyor/forklift-controller/pkg/controller/provider/model/ovirt"
 )
 
 //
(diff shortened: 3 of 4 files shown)


pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -58,6 +59,15 @@ const (
 	Unknown        = "unknown"
 )
 
+// CPU Pinning Policies
+const (
+	None           = "none"
+	Manual         = "manual"
+	ResizeAndPin   = "resize_and_pin_numa"
+	Dedicated      = "dedicated"
+	IsolateThreads = "isolate_threads"
```
liranr23, line 68: For manual, resize and isolated - a PR for validation should be added where a warning will be shown that it won't be respected.
As for the dedicated - we may add a warning as well about the requirement enabling `CPU manager`. - that's an open question.

pkg/controller/plan/adapter/ovirt/builder.go
```diff
@@ -358,6 +368,10 @@ func (r *Builder) mapCPU(vm *model.Workload, object *cnv.VirtualMachineSpec) {
 		Cores:   uint32(vm.CpuCores),
 		Threads: uint32(vm.CpuThreads),
 	}
+	if vm.CpuPinningPolicy == Dedicated {
+		object.Template.Spec.Domain.CPU.DedicatedCPUPlacement = true
```
liranr23, line 372: Although I added only this value, we can also isolate the QEMU emulator. We also may combine it with QoS for CPU and/or [`CPUFeature`](https://kubevirt.io/api-reference/master/definitions.html#_v1_cpufeature).

pkg/controller/plan/adapter/ovirt/builder.go
nyoxi: If the migration blocks without CPU manager and we know that it is not enabled on destination then I would go with an error and not just warning.

liranr23: OK, one way to do it in my opinion is listing the hosts and check for the label `cpumanager=true`. i don't know if that's enough. also, i'm unsure if it's possible to do on validation before starting migration or only blocking during migration.

ahadas: I think you're talking about two different things - Tomas talks about the migration being block, as if the VM is rejected in this case, you Liran talks about the inability to schedule the VM. is that right?

liranr23: No. Because you can't schedule the VM you can't proceed in the migration flow. Therefore, it actually blocks the migration.

ahadas: so you say that the vm starts during the migration? the VM was down at the end of the migrations I did and I don't remember seeing it up..

liranr23: yes, it's the `ScheduleVM` step. This triggers the copy disks. On `CopyDisks` we check the progress.

ahadas: ok, in that case you both were talking about the same thing indeed - but that makes it simpler. if we don't check for other things that would "ensure" the vm can be scheduled on any of the hosts, I don't see why doing it here - forklift should handle that in a more generic way (not on a per-property basis) I guess

rokkbert: > For manual, resize and isolated - a PR for validation should be added where a warning will be shown that it won't be respected. 

So this PR is only supposed to set the "dedicated" flag, if that is what the source VM wants, and just ignore all other values, right?

liranr23: yes

nyoxi: Ok, so if we have same problem with other parameters (mem, cpu,..?) and we don't error there then yes, let's keep the behavior consistent.

nyoxi (approved):

pkg/controller/plan/adapter/ovirt/builder.go
liranr23: based on the offline discussion, we shouldn't check `dedicated` as a special case. CNV cluster should have node with `cpumanger` enabled, and it's the same as broader problem of not having enough resources(like memory and so on).
we can add a warning/info about it, but that also open to discussion.
the relevant validation will be executed in another PR to forklift-validation.

liranr23:
Validation PR: https://github.com/konveyor/forklift-validation/pull/49

pkg/controller/plan/adapter/ovirt/builder.go
ahadas: IIRC we don't isolate the emulator thread or change QoS/CPU features in ovirt when setting the dedicated CPUs policy, no?

pkg/controller/provider/model/ovirt/model.go
```diff
@@ -131,6 +131,7 @@ type VM struct {
 	CpuCores                    int16            `sql:""`
 	CpuThreads                  int16            `sql:""`
 	CpuAffinity                 []CpuPinning     `sql:""`
+	CpuPinningPolicy            string           `sql:""`
```
ahadas, line 134: that's a relatively recent addition to the api - does the "infrastructure" handle it well when it's missing?

ahadas (commented):
overall looks good, we're aware of gaps between kubevirt and ovirt when it comes to cpu pinning - we don't need to solve them here but to map between features and I think it's done well here

pkg/controller/plan/adapter/ovirt/builder.go
liranr23: > IIRC we don't isolate the emulator thread or change QoS/CPU features in ovirt when setting the dedicated CPUs policy, no?

yes, we don't change any of them. the only close thing to the emulator is what we take for VDSM. it's not the same of course.

pkg/controller/provider/model/ovirt/model.go
liranr23: i'm pretty sure it is. for example you may have or not the `CpuAffinity` above. Also from my other PR, I don't remember exactly if it shows empty or nil.

ahadas (approved):
lgtm under the assumption that the case in which the cpu pinning policy property is not reported is verified

liranr23:
> lgtm under the assumption that the case in which the cpu pinning policy property is not reported is verified

based on checking other part, when a field wasn't available from the API, I had: `Required value` error(empty string). Therefore I think we are good to go

ahadas:
> > lgtm under the assumption that the case in which the cpu pinning policy property is not reported is verified
> 
> based on checking other part, when a field wasn't available from the API, I had: `Required value` error(empty string). Therefore I think we are good to go

not sure I understand - so if we work against RHV version that doesn't report the CPU pinning field we'll get an error that says "Required value"? I wouldn't expect to get an error in this case but treat that as 'none'..

ahadas:
ok, that's good. but while speaking on that, I'd check if we have a good place to replace empty string to 'none', just to prevent someone adding code that enumerates the values that are listed and otherwise raise an error

PR #460, opened 2022-08-23
Add admission webhook controller

mnecas:
The admission webhook can mutate or validate requests for creation of
Kubernetes object. This patch adds the validation logic of the webhook.
When the user/UI creates a secret with specific labels it will validate
the connection specified in the secret.

### DIFF
--- a/pkg/webhook/validator.go
+++ b/pkg/webhook/validator.go
@@ -1,0 +1,74 @@
+package webhook
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	liberr "github.com/konveyor/controller/pkg/error"
+	"github.com/konveyor/controller/pkg/logging"
+	api "github.com/konveyor/forklift-controller/pkg/apis/forklift/v1beta1"
+	"github.com/konveyor/forklift-controller/pkg/controller/provider/container"
+	core "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	// Name.
+	Name = "admission-webhook"
+)
+
+// Package logger.
+var log = logging.WithName(Name)
+
+type Handler struct {
+	Client  client.Client
+	decoder *admission.Decoder
+}
+
+//
+// Build custom provider with parameters form secret
+func buildProvider(provider *api.Provider, providerType *api.ProviderType, secret *core.Secret) {
+	provider.Spec.URL = string(secret.Data["url"])
+	provider.Spec.Type = providerType
+	provider.Name = secret.Name
+	provider.Namespace = secret.Namespace
+}
+
+//
+// Admission webhook handler
+// Tests the connection to the provider before creating the secret
+func (a *Handler) Handle(ctx context.Context, req admission.Request) admission.Response {
+	secret := &core.Secret{}
+	err := a.decoder.Decode(req, secret)
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+	// Get the provider type on which start the tests
+	createdForProviderType, ok := secret.GetLabels()["createdForProviderType"]
+	if !ok {
+		return admission.Errored(http.StatusBadRequest, liberr.New("The label 'createdForProviderType' is not set on secret"))
+	}
+	provider := &api.Provider{}
+	providerType := api.ProviderType(createdForProviderType)
+	buildProvider(provider, &providerType, secret)
+
+	collector := container.Build(nil, provider, secret)
+	if collector == nil {
+		return admission.Errored(http.StatusBadRequest, liberr.New(fmt.Sprintf("Incorrect 'createdForProviderType' value. Options %s", api.ProviderTypes)))
+	}
+	log.Info("Starting provider connection test")
+	err = collector.Test()
+	if err != nil {
+		return admission.ValidationResponse(false, err.Error())
+	}
+	log.Info("Provider connection test passed")
+	return admission.ValidationResponse(true, "Passed the validaiton")
+}
+
+// Injects the decoder of the API request
+func (p *Handler) InjectDecoder(d *admission.Decoder) error {
+	p.decoder = d
+	return nil
+}
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -37,6 +37,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
 //
@@ -80,6 +81,8 @@ func main() {
 	log.Info("setting up manager")
 	mgr, err := manager.New(cfg, manager.Options{
 		MetricsBindAddress: Settings.Metrics.Address(),
+		CertDir:            "/var/run/secrets/forklift-admission-webhook-serving-cert/",
+		Port:               8444,
 	})
 	if err != nil {
 		log.Error(err, "unable to set up overall controller manager")
@@ -87,6 +90,10 @@ func main() {
 	}
 
 	log.Info("Registering Components.")
+
+	if Settings.Role.Has(settings.MainRole) {
+		mgr.GetWebhookServer().Register("/webhooks/provider", &admission.Webhook{Handler: &webhook.Handler{Client: mgr.GetClient()}})
+	}
 
 	// Setup Scheme for all resources
 	log.Info("setting up scheme")


jortel:
This deviates with the deliberate (konveyor) design pattern of: 1) controllers performing validation; 2) Representing resource errors as conditions.  Also, admission webhooks adds a bit of complexity by adding an additional failure workflow for the caller.  The webhook can only perform fairly limited validation (and not worth it, IMHO).  In this case, the secret credentials can be invalid for a number of reasons.  The most likely reason is user-provided invalid credentials which should not be validated in the webhook but rather initially/periodically by the controller.  Credentials can expire and/or be updated in the provider itself.
Another drawback with webhooks (in general) is that resources can only be created when the hook (controller) is running.

Webhooks can make working with resources painful and would recommend against unless absolutely necessary.

michalskrivanek:
I concur. We did discuss to block only in case of invalid credentials (i.e. provider is reachable, just auth fails) which should cover the case of creating resources up front when provider doesn't exist/runs. So purely from UX perspective I think it is more or less ok. 
But all the other caveats and complexities indeed doesn't seem to justify introducing it. (Martin can add about the problem with TLS being enforced for hooks, that just complicates non-OCP and dev use cases)

mnecas:
> I concur. We did discuss to block only in case of invalid credentials (i.e. provider is reachable, just auth fails) which should cover the case of creating resources up front when provider doesn't exist/runs. So purely from UX perspective I think it is more or less ok.

This will require something similar as in https://github.com/konveyor/forklift-controller/pull/447, problem is that the test of connection does not return the status code and also some of the provider libraries don't return the status code. So probably will need to rewrite some of the tests to basic HTTP requests.

> But all the other caveats and complexities indeed doesn't seem to justify introducing it. (Martin can add about the problem with TLS being enforced for hooks, that just complicates non-OCP and dev use cases)

Yeah, this is my current problem... the webhooks need to have the TLS enabled currently we are using the special openshift CA injection inside the operator, but this does not work on the Kubernetes. All internal communication in Kubernetes is unsecure and when we try to enable it in the operator the install fails for missing resources that are generated by the openshift tool. So I did some research and some kubevirt operators use the `cert-manager.io` to handle all internal certificates. 
I want to enable this feature for all our applications so we can use the TLS even on Kubernetes.

----


The only alternative to the webhooks I can think of is to stop the reconciliation.
For this, I still would need to rewrite some of the code in the #447 to have the status code. I would need the https://github.com/konveyor/forklift-controller/pull/453 to stop the reconciliation right after the connection test or some other alternative like defer which would set the req to 0.
But I had problems when I was testing the stop of reconciling I found out that it was not 100% reliable. I could see that the reconcile stopped with req 0 but still started another reconcile. Usually, it ended after two or three reconciliations. (It is possible that something changed the provider and triggered the reconciliation, but I did not change anything in the provider myself). This could cause a lock of the account also really quickly.
For me, the webhook seemed more reliable from this perspective. 
```
{"level":"info","ts":1661420022.180439,"logger":"provider|gsb6b","msg":"Condition added.","provider":"konveyor-forklift/test","condition":{"type":"ConnectionAuthFailed","status":"True","reason":"Tested","category":"Critical","message":"Connection auth failed, failed: Unauthorized","lastTransitionTime":null}}
{"level":"info","ts":1661420022.1893876,"logger":"provider|gsb6b","msg":"Waiting connection tested or inventory created.","provider":"konveyor-forklift/test"}
{"level":"info","ts":1661420022.1894202,"logger":"provider|gsb6b","msg":"Reconcile ended.","provider":"konveyor-forklift/test","reQ":0}
{"level":"info","ts":1661420022.1894884,"logger":"provider|vjhbz","msg":"Reconcile started.","provider":"konveyor-forklift/test"}
[GIN] 2022/08/25 - 11:33:47 | 200 |      642.42µs |       127.0.0.1 | GET      "/providers?detail=1"
{"level":"info","ts":1661420027.2874188,"logger":"provider|vjhbz","msg":"Waiting connection tested or inventory created.","provider":"konveyor-forklift/test"}
{"level":"info","ts":1661420027.28746,"logger":"provider|vjhbz","msg":"Reconcile ended.","provider":"konveyor-forklift/test","reQ":0}
``` 
Do you know what could cause this issue? 
Or do you have any alternative ideas?

ahadas:
> The webhook can only perform fairly limited validation (and not worth it, IMHO)

I agree with the first part but not with the last part (that it doesn't worth it)

The idea is to prevent the system from having a desired state that cannot be achieved.
Let's take kubevirt for instance, we could have said the same thing - let all VMIs be added and later on controllers will validate them and set conditions to reflect invalid things (virt-controller also operates this way).
But then if I post a VM with invalid or missing configuration and I don't get immediate feedback - I, as a user, need to wait for the entity to be added, for the condition to be set, then check what's that condition and edit the entity. A better user experience is to tell the user immediately what needs to be changed - that's why we've made the effort of adding such validations to the admission webhooks in kubevirt.

Concretely about this case, @mansam raised a valid point by saying that an invalid credentials doesn't mean the provider is invalid and therefore we changed this to use the secret. If the secret is invalid, we can, as @jortel wrote above, assume that it typically means the user specified invalid input - and then we better reject it. For the other cases, sure, we can stop the reconcile operation when we fail to authenticate but those case are much less likely to happen - the aforementioned scenario is much more likely to happen and affects user experience

ahadas:
> We did discuss to block only in case of invalid credentials (i.e. provider is reachable, just auth fails) which should cover the case of creating resources up front when provider doesn't exist/runs

I don't think that's what @jortel referred to above, I think he meant that when forklift-controller doesn't run, it wouldn't be possible to post entities that are to be validated by the controller (as opposed to adding them and later the controller would validate them and add the relevant conditions if needed).
About that, yes, I agree that it would be better to have a separate pod that handles admission webhooks, like virt-api in kubevirt. The problem is that when we talked about it we saw that every pod is represented as a different repository on konveyor so we'd first want to reorganize the repository of forklift to include all the backend parts and then it would be easier to add something like forklift-api that would handle just that.

ahadas:
> > But all the other caveats and complexities indeed doesn't seem to justify introducing it. (Martin can add about the problem with TLS being enforced for hooks, that just complicates non-OCP and dev use cases)
> 
> Yeah, this is my current problem... the webhooks need to have the TLS enabled currently we are using the special openshift CA injection inside the operator, but this does not work on the Kubernetes. All internal communication in Kubernetes is unsecure and when we try to enable it in the operator the install fails for missing resources that are generated by the openshift tool. So I did some research and some kubevirt operators use the `cert-manager.io` to handle all internal certificates. I want to enable this feature for all our applications so we can use the TLS even on Kubernetes.

+1
we shouldn't blindly follow everything that kubevirt did, there are some things we would probably have done differently now if we started to code kubevirt from scratch. but it is a great reference - so many good people worked on it, there's a lot we can take from it.
 
> The only alternative to the webhooks I can think of is to stop the reconciliation. For this, I still would need to rewrite some of the code in the #447 to have the status code. I would need the #453 to stop the reconciliation right after the connection test or some other alternative like defer which would set the req to 0.
> 
> But I had problems when I was testing the stop of reconciling I found out that when user creates the provider with a secret it starts two reconciles. This is caused by the watch of the provider CR and Secret so both watches might get triggered starting two reconciliations and one waiting after the other finishes.
> 
> https://github.com/konveyor/forklift-controller/blob/8d67009dee1bfe152213a9df1948f1f59449fe69/pkg/controller/provider/controller.go#L101-L114
> 
> For me, the webhook seemed more reliable from this perspective.
> 
> What do you think? Do you have another idea of how we could solve the lock of the account issue?

I think we better have both but rejecting invalid credentials immediately is the more important part.

kaovilai:
Please keep in mind when adding admission webhook controllers that disallow certain resource creation, It could affects backup/restore operations. Please document accordingly.

See: https://github.com/knative/serving/issues/10926 for example issues.

ahadas:
@kaovilai I don't know how relevant is backup/restore to forklift - we're talking about a relatively low number of source/destination environments for which we prefer to pull the current state rather than operating on an old state..

But let's say it is relevant, we're talking about the following scenario:
1. We have an environment where we have a provider with valid credentials
2. Backup is taken
3. When restoring the backup we reject the provider because the credentials are invalid

It's very unlikely but it's possible and what we'll see is that the provider is added and its secret is rejected by the admission webhook, I don't see a problem with that - the user would anyway need to modify the provider and fix its credentials

kaovilai:
> I don't see benefit in adding incorrect secret in this case

agree.
But people Velero users have backup and restore many kind of things and we have encountered many occurences of admission webhooks denying creation of an object resulting in restore "failure" status by velero.

Fail is an alarming word to users and their first reaction is that "Velero doesn't work". We'd then have to walk them through how restore create objects and that creating such objects can someitmes be blocked. by admission/validating webhooks. It would be extremely helpful to document outside admission webhook errors so that users are aware.

kaovilai:
Basically the ask is to document all kinds of errors and how to resolve them somewhere other than just in .go code.

kaovilai:
A troubleshooting doc would be a fitting place for this.

ahadas:
> agree. But people Velero users have backup and restore many kind of things and we have encountered many occurences of admission webhooks denying creation of an object resulting in restore "failure" status by velero.

Of course, it's a common pattern, also [documented in k8s docs](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/) so I'm not surprised you see this a lot :)
Sure, point taken about the need to document that

pkg/webhook/validator.go
```diff
@@ -0,0 +1,74 @@
...
+func (a *Handler) Handle(ctx context.Context, req admission.Request) admission.Response {
+	secret := &core.Secret{}
+	err := a.decoder.Decode(req, secret)
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+	// Get the provider type on which start the tests
+	createdForProviderType, ok := secret.GetLabels()["createdForProviderType"]
+	if !ok {
+		return admission.Errored(http.StatusBadRequest, liberr.New("The label 'createdForProviderType' is not set on secret"))
```
ahadas, line 51: this should be dropped - we still want to accept secrets without this label, and validate those that include the label
I played with this change locally and besides that it works well - I'll include this in #457  along with the 'forklift-api' pod

mnecas: This might be a problem because the secrets don't specify for which provider type they are. We can choose by the data fields. For example, only rhv has cacert or vmware has thumbprint but this might not be 100% reliable. Or reading the url parsing to the provider type according to it.

ahadas: I'm ok with checking the provider type the way you do it now, if the label exists
but the api server can get other secrets, unrelated to MTV providers, that shouldn't be rejected

mnecas: Those will not trigger the webhook please see [1]. Only secrets with the label `createdForResourceType: providers` will trigger the webhook.
[1] https://github.com/konveyor/forklift-operator/pull/247/files#diff-0ae11594eeec8b4ee7eebf3797d21139a635edf688a55164f674355f49c47ff3R13-R16

ahadas: ack, you see - that's what I don't like in this repo-per-component structure - that's it's easy to miss such things.. another argument for having an operator that is closer to the "backend" components..

ahadas:
as we a talked on the demo session, it needs to get in with an additional pod for handling the webhooks - done in kubev2v/forklift so closing this one

PR #463, opened 2022-09-05
Better handling of external tools in Makefile

nyoxi:
The PR enhances handling of external tools (podman, controller-gen) in Makefile. Changes are following:

* Errors are silenced when looking up tools because they are expected.
* Fix pulling of controller-gen if it is not installed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,12 @@ else
     CONTAINER_CMD = docker
 endif
 
+ifeq (, $(shell which controller-gen))
+    CONTROLLER_GEN := $(GOBIN)/controller-gen
+else
+    CONTROLLER_GEN := $(shell which controller-gen)
+endif
+
 ci: all
 
 all: test manager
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -69,14 +77,6 @@ docker-build:
...
-ifeq (, $(shell which controller-gen))
-	@{ \
-	go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.6.2 ;\
-	}
-CONTROLLER_GEN=$(GOBIN)/controller-gen
-else
-CONTROLLER_GEN=$(shell which controller-gen)
-endif
+$(CONTROLLER_GEN):
+	go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.6.2
```
mnecas, line 82: Why have you renamed it from `controller-gen` to `$(CONTROLLER_GEN)`?
Now the user would need to specify the full path with make like `make /usr/bin/controller-gen` when he would need to install it, but that can change.

nyoxi: Ok, updated. I didn't think it is necessary but I have created also a phony target `controller-gen`. It allows one to install `controller-gen` simply with `make controller-gen` if one wants. And hopefully it will also make the Makefile dependencies little bit easier to follow.

ahadas:
@nyoxi still makes sense to post on kubev2v/forklift?

